PR #5, opened 2017-04-04
Replace the reddit.com/place palette with the superior db16 palette.

DarkLinkXXXX:
As demonstrated in [this thread on PixelJoint](http://pixeljoint.com/forum/forum_posts.asp?TID=12795), db16 is a really nice color palette. Better than the reddit palette, methinks.

Palette infographic: 

Cool art made with it (not mine):


Sbrown-WilliamsCrusher:
This is what the Current board would look like if palette was changed [img](https://gyazo.com/466ca01dbc3563475bebb6535ae57b21 )
Thanks to Jaron780 from discord

DarkLinkXXXX:
@Sbrown-WilliamsCrusher I see absolutely nothing wrong with that!

In all seriousness though, we probably shouldn't change the colorscheme without clearing the board, which probably shouldn't be done just to change the color scheme. Also, some may disagree, but I'd switch black and white. I didn't realize the whole board would be painted with the first color.

YenTheFirst:
That mockup isn't quite representative, since the two palettes are in different orders. Here's a more representative mockup:

current:
pxlspace_curr

dba16:
pxlspace_dba16

this reorder was just done by eye, there might be a more optimal reorder, and some of the artwork certainly would have made different choices with a different palette. The biggest notable difference is that the current palette has two subtly different whites, where dba16 has a white and a grey.

Sbrown-WilliamsCrusher:
The PR was just pulled to a dev environment with no reordering that's why it looks odd. I did reorder them but didn't get a chance to screenshot it. Guess I don't have to anymore, thanks @YenTheFirst

gunnerwolf:
Not gunna happen. We might *add* colors in the future, but we have no intention of ever *replacing* the current core 16 pallette

PR #8, opened 2017-04-04
Add timer text to page title

### LINKED ISSUE #6: Add timer in page title
So we can see time left from another tab, or when minimized.

EdenSG:
Resolves #6

Sorry for the messy three commits instead of one :P


EdenSG:
Fixed by e53a1b7c8d21568c7a8c59a0926060dba43ce213

PR #9, opened 2017-04-04
Added Shutdown command

thamstras:
Added a shutdown command and some graceful shutdown handling.
It sends a WS broadcast, but doesn't do any client handling beyond that.


Sorunome:
way too old anyways, re-open if you want to update with a newer version thing

PR #10, opened 2017-04-04
Use WebSocket for users

qwirk:


resources/public/app.js
```diff
@@ -257,7 +260,7 @@ window.App = {
 
         this.socket = ws;
     },
-    initUsers: function () {
+    /*initUsers: function () {
```
EdenSG: why not just remove this code - it's git, the original code is preserved

PR #30, opened 2017-04-11
get mobile working

Sorunome:
works at leats in firefox mobile, as the events were quitemessed up it
should work quite universal

And i have no idea why it says "can't automatically merge" and yet is abgle to show all the changes....


Sorunome:
The newest sneak-commit of ske breaks the design, though, so that the login buttons are hidden. This only fixes pixel-placing with touch

PR #31, opened 2017-04-12
Add ctrl-click functionality to get a URL at that position

Eearslya:
Take 2, since Git didn't want to cooperate.

Ctrl-click anywhere on the canvas to get an alert box with a URL for the direct link to that pixel.

Rudimentary setup for now, because browsers purposefully like making it difficult for JavaScript to mess with the clipboard.

### DIFF
--- a/resources/public/app.js
+++ b/resources/public/app.js
@@ -200,12 +200,19 @@ window.App = {
         var downX, downY;
 
         this.elements.board.on("pointerdown mousedown", function (evt) {
+            if (evt.originalEvent.ctrlKey) {
+                var pos = this.screenToBoardSpace(evt.clientX, evt.clientY);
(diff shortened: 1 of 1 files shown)


jay-bricksoft:
You could take a look at browserify and other polyfills for clipboard. Will have a look at it when I'm home

Netox005:
How about changing the path of the URL on the address bar? The code would be something like `window.history.pushState('', document.title, `/?x=${getXPosition()}&y=${getYPosition()}`);`

jasperandrew:
Oh, cool. Keep up the good work, man, less for me to do the better!

Sorunome:
kinda unneeded by now as the URL auto-updates all the time

PR #32, opened 2017-04-12
Fix the bar sliding up on an invalid captcha

Sorunome:
change all positions to fixed
This fixes the bar sliding up on an invalid captcha

Stuff like this: https://puu.sh/vhY6v.png


xSke:
Doesn't this break mobile?

Sorunome:
nope, tested it on mobile. In our case `position:fixed` vs `position:absolute` shouldn't even be a difference as the webapp doesn't scroll

PR #37, opened 2017-04-14
get MS edge to work

Sorunome:


resources/public/app.js
```diff
@@ -118,6 +144,19 @@ window.App = {
         }
 
         ctx.putImageData(id, 0, 0);
+        if (this.use_js_resize) {
+            var ctx2 = $(".board-container canvas")[0].getContext("2d");
+            ctx2.canvas.width = window.innerWidth;
+            ctx2.canvas.height = window.innerHeight;
+            ctx2.mozImageSmoothingEnabled = false;
```
Netox005: why use all the prefixes if only Edge is affected?

Sorunome: I wanted to make `use_js_resize` to work with potentially any browser we'd happen to trip over on our way, and thus I just added all the prefixes

PR #41, opened 2017-04-14
Change App.saveImage

hammer065:
This version of App.saveImage is simpler and more cross-browser compatible than the old code.


hammer065:
@gunnerwolf Change made, please review
a.remove is wrapped since I cannot find an official documentation to this function but browsers seem to support it, thats why I didn't implement it in the beginning.

PR #42, opened 2017-04-15
Correct JSON in README.md

hammer065:


gunnerwolf:
The config file is not JSON, it just uses a similar formatting style.

xSke:
Yes, it's actually [HOCON](https://github.com/typesafehub/config/blob/master/HOCON.md), a superset of JSON.

PR #49, opened 2017-04-16
added templates, so far only via URL parameters

Sorunome:
URL parameters are:

* `template` - URL of the image
* `ox` - x-coordinate of the image (optional, default: 0)
* `oy` - y-coordinate of the image (optional, default: 0)
* `oo` - opacity of the iamge (optional, default: 0.5)
* `tw` - width of the template (optional, default: image width)

### DIFF
--- a/resources/public/app.js
+++ b/resources/public/app.js
@@ -68,6 +68,14 @@ window.App = {
         loginOverlay: $(".login-overlay"),
         userInfo: $(".userinfo")
     },
+    template: {
+        use: false,
+        url: '',
+        x: 0,
+        y: 0,
+        width: -1,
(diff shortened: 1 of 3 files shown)


Sorunome:
Don't merge yet, i'm doing some legit work on it that i just came up with

Sorunome:
There you go, now you can merge. I added a function `App.updateTemplate` which works like the following:

```
App.updateTemplate({
    use: true/false,
    url: 'string',
    x: number,
    y: number,
    opacity: float,
    width: number
})
```

Stuff is passed as an object so that you can easily just omit any parameter that you don#t want to change

PR #54, opened 2017-04-17
fix timer countdown on firefox mobile

Sorunome:
for some reason it doesn't always like setInterval, calling
setTimeout repetitively seems way more reliable


Sorunome:
Some of my thinking was backwards, this can be done way more efficiantly, will create a new PR then

PR #57, opened 2017-04-18
Adds the ability to click on the Pxls notification

brandonsturgeon:
This is really bad of me to do this, but I actually don't have access to a good Java development platform right now.

I *believe* this change will work, but I'd need someone to quickly test it for me  :(

Here's what it should do:

1. It should use the favicon as the icon for the notification
2. It should allow the player to click on the notification to be taken to the current pxls.space tab/window, and then close the notification.

I would really appreciate it if someone could give it a quick test and let me know if it works as intended!

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -717,6 +717,13 @@ if (ms_edge) {
             alert.find(".text").text(message);
             alert.fadeIn(200);
         },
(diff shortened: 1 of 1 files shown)


brandonsturgeon:
I will also have to make sure that this works with https://github.com/xSke/Pxls/pull/52

PR #60, opened 2017-04-18
Makes the notification close when clicked

brandonsturgeon:
This is a followup to my last pull request: https://github.com/xSke/Pxls/pull/57

This should make it so when the notification is clicked, it actually closes the notification too.

I can't actually test this locally, but here's a jsfiddle demonstrating what this code should do. (It's the same code): https://jsfiddle.net/brandonsturgeon/puwpwf36/

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -739,6 +739,7 @@ window.App = (function () {
                 notification.onclick = function() {
                     window.focus();
                     this.cancel();
+                    notification.close()
                 };
             } catch(e) {
                 console.log("No notifications available!");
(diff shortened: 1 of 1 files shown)


Sorunome:
So I was asked on discord to test this.

While I cannot verify its functionality because of my custom archlinux setup (all notifications always close when clicking, no matter what) I can still give my thoughts on it:

It seems like using `this.close()` is the correct way to close it, just like it was done here. The code also doesn't yield any runtime errors so there is no reason that I can see why it shouldn't work. Here are the docs for anyone: https://developer.mozilla.org/en-US/docs/Web/API/Notification/close

So, if it was up to me, i'd say merge it ^.^

gunnerwolf:
I trust sorunome's input, merging.

PR #62, opened 2017-04-20
Client-side JS restructure

Sorunome:
Obviously do not merge this PR yet, it is here so that others can see what kind of restructure is happening.

Here is the current documentation on how the client-side will be structured then: https://gist.github.com/Sorunome/fe1519d8cd297c43c7e5b7264e69385d

Objects might adapt slightly when actually implementing if I notice that things work out way better if outlined a bit differently


Sorunome:
What works so far:
* client.side rendendering using css, `use_zoom` and `use_js_render`
* templates, and placing pixels under them
* captchas
* like everything non-mod related

Tested browsers:
* firefox
* chromium
* opera
* vivaldi
* ms edge
* mobile firefox
* ios safari

What still needs to be done:
* mod / `admin.js` stuff

Sorunome:
OK, so this should be ready to merge now, I tested it in quite some browsers as listed above. Someone else please do tests, too, because, well, this *is* a large client-side restructure and with such things errors just happen.

This also implements #63

PR #63, opened 2017-04-21
Close Socket when the site unloads

hammer065:
Close Socket when the site unloads, so the Server doesn't retransmit the first packet it tries to send after the Socket doesn't listen anymore (which it does)


hammer065:
Implemented in #62 ec3e3cd5060d155d8cddbb5efa8ecf4032cae32c
Closed.

PR #70, opened 2017-05-04
Adds rollback after a user gets banned

LittleEndu:


xSke:
gonna check a bunch of stuff, then likely merge, thanks!

PR #71, opened 2017-05-06
Better rollback

LittleEndu:
This one uses previous pixel's ID instead of just saving previous color. This means we get more control over what we rollback and what we undo. You can read more about it in the code. The comments should explain it.

### DIFF
--- a/src/main/java/space/pxls/App.java
+++ b/src/main/java/space/pxls/App.java
@@ -6,6 +6,8 @@ import com.typesafe.config.ConfigFactory;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import space.pxls.data.DBPixelPlacement;
+import space.pxls.data.DBRollbackPixel;
 import space.pxls.data.Database;
 import space.pxls.server.Packet;
 import space.pxls.server.UndertowServer;
@@ -20,10 +22,7 @@ import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Scanner;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.TimeUnit;
 
 public class App {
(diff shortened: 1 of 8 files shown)


src/main/java/space/pxls/App.java
```diff
@@ -175,25 +174,52 @@ public static int getPixel(int x, int y) {
...
+        for (DBRollbackPixel rbPixel : pixels) {
+            //This is same for both instances
+            //  putPixel() logs and updates the board[]
+            //  forBroadcast.add() adds the pixel and later broadcasts it via websocket
+            //  putRollbackPixel() adds rollback pixel to database (TABLE pixels) for undo and timelapse purposes
+            if (rbPixel.toPixel != null) { //if previous pixel (the one we are rolling back to) exists
+                putPixel(rbPixel.toPixel.x, rbPixel.toPixel.y, rbPixel.toPixel.color, who, false, "(rollback)", false);
+                forBroadcast.add(new Packet.ServerPlace.Pixel(rbPixel.toPixel.x, rbPixel.toPixel.y, rbPixel.toPixel.color));
+                database.putRollbackPixel(who, rbPixel.fromId, rbPixel.toPixel.id);
+            }else{ //else rollback to blank canvas
```
Sorunome: Format! Should be `} else {`

src/main/java/space/pxls/user/UserManager.java
```diff
@@ -128,21 +128,21 @@ public void banUser(User user, long timeFromNowSeconds, String reason) {
     }
 
     public void banUser(User user, long timeFromNowSeconds) {
-        banUser(user, timeFromNowSeconds, "", 0);
+        banUser(user, timeFromNowSeconds, "", 24*3600);
```
Sorunome: Please have this default to `0` because `banUser` is only called on time-based bans and they aren't as heavy as permabans

LittleEndu: Shouldn't it still default to some non-zero value? Like 1 hour

Sorunome: hm yeah one hour sounds good, too. Mind chaning that here, too? https://github.com/xSke/Pxls/blob/master/resources/public/admin/admin.js#L70

so that instead of `time ? 0 : 24` it says `time ? 1 : 24`

LittleEndu: Actually it doesn't even matter because L131 calls L126 and that already has default time.

Sorunome: lol mind changing that defautl time to one hour, too? ^.^

PR #72, opened 2017-05-07
Change to broadcast_noshadow()

LittleEndu:


LittleEndu:
Going to commit with right message :P

PR #81, opened 2017-05-10
Multi threaded rollback and undo

LittleEndu:


LittleEndu:
Implemented via worker threads

PR #87, opened 2017-06-24
Ability to lock the canvas

Pitu:
When placing multiple pixels very quickly you often encounter an issue were the you start dragging around instead of placing pixels. When working on big drawings this is specially annoying since instead of placing you end up moving the canvas away. 

This PR proposes a way to prevent this by pressing L to toggle the locked state and preventing the user to drag the canvas around or zoom into it.

### DIFF
--- a/resources/public/index.html
+++ b/resources/public/index.html
@@ -76,52 +76,18 @@
             <div class="close">Close</div>
             <div>
                 <h1>Welcome!</h1>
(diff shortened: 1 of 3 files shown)


Pitu:
Now it also supports moving the canvas around by pressing the arrow keys or WASD. On of our users requested this feature so I added it here as well.

PR #96, opened 2018-01-25
 Move stacker count inside #cursor

netux:
I didn't really like that the stacker count was stuck in a bubble at the bottom left of the screen.

This way the counter is not only more visible, but hints more on how the stacking mechanic works.

demo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ target/
 run/
 logs/
 backups/
+.settings/
+.classpath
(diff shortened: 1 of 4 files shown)


netux:
Changes branched off this branch were already merged.

TheRebelG0d:
Merged at bc6db9fc83e7ec3f22bd886d380d7b58538ad564

PR #98, opened 2018-04-03
Added gradle support and some other improvements

Giszmo:
This PR is just some cleanup in the process of getting familiar with the project. I prefer working with gradle - but didn't get fat jars to work yet. But `./gradlew run` works. `./gradlew test` not yet.

Lint convinced me to dispose of more stuff than what was good for the project, so there is some undoing going on in the commits ... but I tamed lint in the process, so others don't fall for it.

I wanted to install and start painting pixels but was a bit disappointed that the project doesn't work out of the box, so I looked into the `AuthService`s and (loally) added a `NoAuthService` and generally made it easier to just offer what is configured (with this PR, only services that have a `key: "not empty"` configured are listed). How would I implement such a `NoAuthService`? Ideally I would have it URL-based, so I can have 3 tabs open, with 3 ad-hoc accounts. Having poked around for 3h today ... I'd really want to know how to skip the oauth stuff for such a purpose. I want to develop with multiple accounts but on an air-gapped machine (no reddit dependency, no dev down-time due to bad internet).


Sorunome:
> I wanted to install and start painting pixels but was a bit disappointed that the project doesn't work out of the box

it used to, but some java or maven update or something broke it just a few days ago :[

Thanks for the PR, i'll take a closer look at it later

Am 3. April 2018 05:05:55 MESZ schrieb Leo Wandersleb <notifications@github.com>:

Giszmo:
> it used to, but some java or maven update or something broke it just a few days ago :[

Apparently 5db22cc94e94870fb637738ee23644e251b4a840 was still good. Is anybody working on reverting/fixing breaking changes?

src/main/java/space/pxls/App.java
```diff
@@ -456,8 +457,7 @@ public static byte getDefaultColor(int x, int y) {
             byte b = raf.readByte();
             raf.close();
             return b;
-        } catch (NoSuchFileException e) {
```
Sorunome: What about the `NoSuchFileException`? a `default_board.dat` doesn't necessarily have to exist

src/main/java/space/pxls/App.java
```diff
@@ -288,7 +289,7 @@ public static int getPixel(int x, int y) {
     }
 
     public static void putPixel(int x, int y, int color, User user, boolean mod_action, String ip, boolean updateDatabase) {
-        if (x < 0 || x >= width || y < 0 || y >= height || (color >= getPalette().size() && !(color == 0xFF || color == -1))) return;
+        if (x < 0 || x >= width || y < 0 || y >= height || color >= getPalette().size() && color != 0xFF) return;
```
Sorunome: Java uses signed ints only, so isn't the byte `0xFF` converted to `-1` when casting to an int?

src/main/java/space/pxls/auth/AuthService.java
```diff
@@ -21,11 +22,14 @@
 
 public abstract class AuthService {
     private String id;
-    private String name = getName();
+    // Gson serialization needs fields, not getters.
+    @SuppressWarnings({"unused", "FieldCanBeLocal"})
+    private String name;
     private transient Set<String> validStates = ConcurrentHashMap.newKeySet();
 
-    public AuthService(String id) {
```
Sorunome: Why are you removing all the `public`, `protected` etc. keywords?

src/main/java/space/pxls/data/DBRollbackPixel.java
```diff
@@ -1,9 +1,5 @@
 package space.pxls.data;
 
-
-/**
- * Created by Endrik on 05-May-17.
- */
```
Sorunome: Why remove comments that correctly credit who wrote pieces of code?

src/main/java/space/pxls/data/Database.java
```diff
@@ -295,7 +295,7 @@ public void addReport(int who, int pixel_id, int x, int y, String message) {
     }
 
     class DatabaseHandle {
-        public DAO dao;
-        public long lastUse;
+        DAO dao;
+        long lastUse;
```
Sorunome: Again, why remove keyword `public`

src/main/java/space/pxls/server/PacketHandler.java
```diff
@@ -163,9 +161,6 @@ private void handleUndo(WebSocketChannel channel, User user, ClientUndo cu){
     }
 
     private void handlePlace(WebSocketChannel channel, User user, ClientPlace cp, String ip) {
-        if (!cp.getType().equals("pixel")) {
-            handlePlaceMaybe(channel, user, cp, ip);
-        }
```
Sorunome: Please do not remove this, even though the function is empty, for some.....reasons (it is used occasionally for anti-bot measures i didn't put on git to hide how exactly it's working)

src/main/java/space/pxls/server/PacketHandler.java
```diff
@@ -247,9 +242,6 @@ private void handlePlace(WebSocketChannel channel, User user, ClientPlace cp, St
         sendCooldownData(user);
     }
 
-    private void handlePlaceMaybe(WebSocketChannel channel, User user, ClientPlace cp, String ip) {
-    }
```
Sorunome: Please do not remove this, even though the function is empty, for some.....reasons (it is used occasionally for anti-bot measures i didn't put on git to hide how exactly it's working)

src/main/java/space/pxls/user/User.java
```diff
@@ -3,19 +3,20 @@
 import io.undertow.websockets.core.WebSocketChannel;
 import space.pxls.App;
 
-import java.time.Instant;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
+import static space.pxls.user.Role.*;
```
Sorunome: Why do you have to do this import and then use `MODERATOR` instead of `Role.MODERATOR`, while the `Role.MODERATOR` clearly shows where that keyword comes from, while your way doesn't?

Sorunome:
> I wanted to install and start painting pixels but was a bit disappointed that the project doesn't work out of the box

https://github.com/xSke/Pxls/blob/master/resources/reference.conf#L87 setting `oauth.useIp` to true will use user auth stuff by IP rather than external auth services

src/main/java/space/pxls/App.java
Giszmo: NoSuchFileException is an IOException. Ignoring the latter makes ignoring the former obsolete.

Sorunome: alright

src/main/java/space/pxls/App.java
Giszmo: lint told me to do this change. I blindly followed and would probably write some code to figure out if I should suppress lint or actually do this change. will look into it later.

src/main/java/space/pxls/auth/AuthService.java
Giszmo: Reducing visibility improves code quality in many ways. I tried to not over-do it, as sometimes methods are in place for later or as part of a library, to be used in other projects as is but else, every public that can be a private is a good thing to change.

src/main/java/space/pxls/data/Database.java
Giszmo: These members are only needed inside the package, so lets not make them visible outside. If somebody needs them outside of the package, chances are he shouldn't and either should move his new class into the package or resolve it in some other way. If after having had a second thought, exposing them makes sense, so he may change them of course.

src/main/java/space/pxls/server/PacketHandler.java
Giszmo: see below ...

src/main/java/space/pxls/server/PacketHandler.java
Sorunome: > Ok, for an open source project not having parts of the source open is a new problem to have for me.

While i have to agree, the amount of people breaking the no-botting rule and peeking into the code how to circumvent everything got insane. Desperate times require desperate measures. I am happy to say that we are running the exact version as here on github since very long again already.

I really wish people would just shut up and follow the rules >.>

src/main/java/space/pxls/data/DBRollbackPixel.java
Sorunome: While authorship is tracked by git, some prefer it this way. So, please don't remove such comments. There is nothing harmful with them, after all

src/main/java/space/pxls/data/Database.java
Sorunome: I have to agree, and i don't have the whole code in my head so i got confused 👍

FlyingSixtySix:
No updates have been made to this in over a year, and requested changes haven’t been resolved. Closed.

PR #100, opened 2018-04-11
Fix "undo" always using default board color when database lookup misses

TheRebelG0d:
This modifies the `pixels` column to track a `lastColor` int. Whenever `App.putPixel` is called, it stores the current pixel from the board array before overwriting with the new color. This stored value is placed in the `lastColor` column. This way, when undoing, the correct color is returned even if there's no pixel history.


Sorunome:
In which cases is this needed?

TheRebelG0d:
Let's say someone has a project that has a lot of red pixels, and the canvases initial background has a spot with red (e.g. sprinkle on the donut canvas). If you were to misplace, then undo on that red background (which has no history in the database), it will turn gray instead of red which isn't actually undoing the mistake. This fixes that. While not a major problem, it is still an annoyance that pops up a lot with the gimmick canvases.

PR #103, opened 2018-05-10
Add illuminati feature from r/place

haykam821:
In the original r/place, there was a button in the corner with an all-seeing eye on it. When pressed, this enabled "Illuminati mode", which automatically centered the board on the newest pixel placed.

With this pull request, the functionality has been brought to Pxls via a toggle in the settings menu.

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -1149,6 +1149,10 @@ window.App = (function () {
                                 self.ctx.fillRect(px.x, px.y, 1, 1);
                                 self.intView[px.y * self.width + px.x] = 0xFF000000 | self.color;
                             });
+
+                            if (ls.get("illuminati")) {
+                                place.centerBoardOn(px.x, px.y);
+                            }
                         });
                     });
                 },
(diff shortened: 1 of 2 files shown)


resources/public/pxls.js
```diff
@@ -1964,6 +1967,11 @@ window.App = (function () {
                     $("#audiotoggle").change(function () {
                         ls.set("audio_muted", this.checked);
                     });
+                    $("#illuminati")[0].checked = ls.get("illuminati");
```
TheRebelG0d: When illuminati is enabled, the checkbox should be checked on page load.

resources/public/pxls.js
```diff
@@ -1679,6 +1679,9 @@ window.App = (function () {
                     socket.on("pixel", function (data) {
                         $.map(data.pixels, function (px) {
                             board.setPixel(px.x, px.y, px.color, false);
+                            if (ls.get("illuminati")) {
+                                board.centerOn(px.x, px.y);
```
TheRebelG0d: If the user enables this and then places a pixel, all open tabs will jump to that pixel, which shouldn't be happening. Probably needs an extra property on the event

haykam821: > Probably needs an extra property on the event

What do you mean exactly?

resources/public/pxls.js
haykam821: What is this problem? For me, the checkbox is checked the same way before and after a refresh.

resources/public/pxls.js
haykam821: Ah, I'll try to implement this but I probably won't succeed since I'm not familiar with Java.

PR #109, opened 2018-07-19
Update donation section

FlyingSixtySix:
Comments out the broken donation button and adds text indicating
unavailability.


haykam821:
I don't see why you're commenting it out. If you want to restore it, you can just revert the commit that makes it unavailable.

PR #111, opened 2018-07-27
Add capability to provide extra data for lookup

haykam821:
This pull request changes the system in which lookup data is provided, allowing developers to add their own data to it, along with making the existing data more capable in the future. The developer guide has been updated to demonstrate how this works, and you can use it in a userscript with `App.lookup.registerHook(hook);`

### DIFF
--- a/developer.md
+++ b/developer.md
@@ -61,3 +61,17 @@ App.ls.get('blah'); // returns the object
 App.ls.remove('blah'); // remove the contents of blah
 App.ls.get('blah'); // not defined, returns undefined
 ```
+
+## Lookup Hooks
+
+It is possible to extend the built-in pixel lookup functionality by registering a hook which has an id, user-facing name, and a function which provides the basic lookup information.
+
+```js
+App.lookup.registerHook({
+    id: "rebel_hook",
+    name: "Rebellious Hook",
+    get: data => `Hehehe... the X is ${data.x} and Y is ${data.y}...`,
+});
+```
+
+Your function's return value is an array of or single DOM element, jQuery object, or string.
--- a/developer.md
+++ b/developer.md
@@ -1,51 +1,59 @@
 # Developer Documentation
+
 What this file is for is to document the different features for developers to either use inside of `pxls.js` when doing dev or what is exposed so that other people can write their own userscripts to do cool stuff!
 
 All the things are called via the object, in case of inside of the `pxls.js` tis is `self`, for the outside it is `App`. This documentation will just use the `App` notation.
 
 ## Templating
+
 Templating is already provided via URL parameters, but you can easily modify it via scripts. For that there is `App.updateTemplate()`. It works by passing a template-object as an argument. A full template object looks like the following:
-```
+
+```js
 App.updateTemplate({
-	use: true, // boolean, true/false
-	url: 'https://example.com/image.png', // string, url of image
-	x: 5, // float, x-position of image
-	y: 42, // float, y-position of image
-	width: 7, // float, width of image, if scaling is desired
-	opacity: 0.5, // opacity, 1 is max, 0 is min
+    use: true, // boolean, true/false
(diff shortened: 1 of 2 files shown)


Sorunome:
What usecase do you see for this?

haykam821:
@Sorunome A few of the main things it can be used for is to let users add custom notes to users, show their Discord tags, and the faction they are aligned with, without having to bake these features into the client itself.

Sorunome:
Sorry, we purposefully cherry-pick which data to display

FlyingSixtySix:
> Sorry, we purposefully cherry-pick which data to display

@Sorunome, what do you mean by this?

Sorunome:
we don't want any additional information to display on user lookup

FlyingSixtySix:
Why restrict users from adding extra client-side information though? This opens up that avenue - while allowing us to keep to our defaults.

haykam821:
A userscript is the user's choice on the client only. No need for main Pxls to support it except for opening up the way to add them more easily.

Sorunome:
Making it easy to display extra stuff like faction would prevent more people from potentially attacking etc. which is already way too little currently.

We discussed this whole thing already for when we discussed if we should add username lookup at all

haykam821:
@Sorunome That faction info is only available via a userscript. If you manage to get that userscript, you're probably in a Discord faction.

Sorunome:
Well, we don't have to support that now

haykam821:
@Sorunome There's an official Discord server where factions are encouraged via the factions channel and diplomat role. Maybe that is what is causing the lack of attacks?

Sorunome:
that is completely different, that is to make it easier to find art others make and join in on it. Now, if a user wants to take a break of that and go over something else....being associated with a faction only hurts.

haykam821:
If a user wants to be associated with the faction... so be it. If they want to have faction info in their lookups... so be it.

Sorunome:
This PR allows others to label other users and have that information very easily accessible, this PR will not be merged.

If you have other feature ideas, maybe ask about them in the issues first before spending time implementing them only for them to not be merged?

haykam821:
How is it very easily accessible? You have to get a userscript to have it show and a whole data set for it to be useful.

FlyingSixtySix:
Keep in mind that this is entirely client-side--what one user does with this has no effect on another user.

Sorunome:
I am.fully aware thjs is client side. Why should we, though, implement an API that goes against some pxls ideas?

FlyingSixtySix:
> that goes against some pxls ideas?
What pxls ideas are we going against?

haykam821:
Which ideas, specifically? I assume you prefer anarchy more than what Place seemed to actually be about:

> There is an empty canvas.
> You may place a tile upon it, but you must wait to place another.
> Individually you can create something.
> Together you can create something more.

Doxylamin:
Sorry for bringing me in now, but my inbox got spammed and I gotta say, i'm with @haykam821 and @FlyingSixtySix on that one.

TheRebelG0d:
This does not go against any "pxls ideals" that I know of... This doesn't return any extra data from the server, so all of the data that could be used by this has already been "cherry picked". All this PR does is allows someone to do the exact same thing as if they were keeping a list of people in notepad and ctrl+f'd every time someone placed a pixel. Some factions already keep a channel dedicated to watching griefers of their project.

I'm with the others on this, I can't really see a reason to block this PR... any chance to reopen and reconsider @Sorunome?

haykam821:
Two of the largest factions have a griefer list in their Discord servers.

* ManePxls: 
ManePxls griefer list
* American Flag in Place: 
AFIP griefer list

These lists have the same amount of accessibility as having it directly provided with a lookup hook.

Sorunome:
@TheRebelG0d i think it is best to move this discussion to our discord

haykam821:
Can we not? I'm banned from the Discord.

PR #112, opened 2018-07-27
Add "Move Template" button on lookups

haykam821:
This allows for managing template positions on mobile without having to use a keyboard or edit the URL parameters.

Example of Move Template button

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -1864,7 +1864,13 @@ window.App = (function () {
                         : ""),
                         $("<div>").addClass("button").css("float", "right").text("Close").click(function () {
                             self.elements.lookup.fadeOut(200);
-                        })
+                        }),
+                        (data ? $("<div>").addClass("button").css("float", "right").text("Move Template").click(function () {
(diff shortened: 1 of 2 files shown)


Sorunome:
maybe chang e the button name to something like "move template here"? i found that labeling a bit confusing at first.

is it possible to hide the button if no template is loaded?

haykam821:
Both of your issues have been addressed.

New preview of Move Template Here button

gunnerwolf:
Is it possible to make the button fit between the other 2 with the longer text if the button styles (padding and margin) are tweaked slightly? Would look a lot cleaner.

haykam821:
Changing the `margin-right` rule on `.button` to `4.5px` solves this issue.

gunnerwolf:
nice one. I won't merge this myself just because I've been inactive on this project for over a year and feel like it's not really my place to merge PRs anymore, so I'll leave that up to @Sorunome, I just wanted to give my $0.02 on the styling of it

Sorunome:
It looks great! I'll merge when on PC, thank you

PR #117, opened 2018-08-01
Convert admin lookup handler to hooks

haykam821:
This pull request removes the admin lookup handler which has code mostly duplicated from the default one, and instead adds in hooks registered and unregistered alongside other admin features. This allows admins to be able to use the hooks feature and gets rid of the need to support two different lookup handlers at once.

### DIFF
--- a/resources/public/admin/admin.css
+++ b/resources/public/admin/admin.css
@@ -2,6 +2,11 @@
 | Admin Box
 \*\\\\\\\\\\\\\\\\\\\\\\\\\\*/
 
+pre {
+    display: inline;
+    white-space: pre-wrap;
+}
+
 .admin {
     position: absolute;
     top: 40vh;
--- a/resources/public/admin/admin.css
+++ b/resources/public/admin/admin.css
@@ -2,11 +2,6 @@
 | Admin Box
 \*\\\\\\\\\\\\\\\\\\\\\\\\\\*/
 
-pre {
-    display: inline;
-    white-space: pre-wrap;
-}
-
 .admin {
     position: absolute;
     top: 40vh;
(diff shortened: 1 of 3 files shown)


FlyingSixtySix:
I'd prefer if the moderator actions were on a single line, and used text instead of emojis.

haykam821:
Here is what they look like now:

image

FlyingSixtySix:
The button font size looks a bit small on my machine.
image

haykam821:
I did that to distinguish between lookups' regular buttons (Report and Close) and hook buttons. Would you prefer them to be normal-sized, like the below example?

Lookup with admin hooks and normal-sized buttons

FlyingSixtySix:
Yes, please.

haykam821:
It has been done.

resources/public/admin/admin.css
```diff
@@ -2,6 +2,11 @@
 | Admin Box
 \*\\\\\\\\\\\\\\\\\\\\\\\\\\*/
 
+pre {
+    display: inline;
```
Sorunome: Why make a `<pre>` element inline, that is what `<code>` or `<span>` elements are for! `<pre>` should stay a block element

resources/public/admin/admin.js
```diff
@@ -297,62 +295,46 @@
...
-                },
+                /**
+                 * Register hooks for admin-specific lookups.
+                 */
                 init: function () {
-                    admin.lookup.registerHandle(self.create);
+                    App.lookup.registerHook({
+                        id: "login",
+                        name: "Login",
+                        get: data => $("<pre>").text(data.login),
```
Sorunome: same down here with using `<pre>` as inline element

resources/public/admin/admin.js
```diff
@@ -297,62 +295,46 @@
...
+                    }, {
+                        id: "user_agent",
+                        name: "User Agent",
+                        get: data => $("<pre>").text(data.userAgent),
```
Sorunome: abd here

Issue #119, opened 2018-08-12
Allow tendril mechanic via option in config


haykam821:
Right now, the tendril mechanic (seen in [pxls.space](http://pxls.space/)'s 6th canvas) has [been commented out](https://github.com/xSke/Pxls/blob/9ce8cf0cad1ff5988003a4df4483fa92b3e224b4/src/main/java/space/pxls/server/PacketHandler.java#L195). This code not being available for use is such a waste. Why not reimplement it under a opt-in configuration parameter accepting a boolean, such as `board.onlyPlaceOffNeighbors`?

PR #132, opened 2018-09-04
Add tendril support

FlyingSixtySix:
This pull request adds support for _tendrils_, or _pixels that may only be placed beside other pixels,_ as a configuration option. This addresses #119.

pxls tendril demo 4

### DIFF
--- a/resources/reference.conf
+++ b/resources/reference.conf
@@ -73,9 +73,6 @@ undo {
 
 selfPixelTimeIncrease: true
 
-// Pixels can only be placed beside one that has already been placed
-tendrils: false
-
 // System host, used for various things
 host: ""
 
--- a/resources/reference.conf
+++ b/resources/reference.conf
@@ -73,6 +73,9 @@ undo {
 
 selfPixelTimeIncrease: true
 
+// Pixels can only be placed beside one that has already been placed
+tendrils: false
+
 // System host, used for various things
 host: ""
 
--- a/src/main/java/space/pxls/server/PacketHandler.java
+++ b/src/main/java/space/pxls/server/PacketHandler.java
@@ -192,25 +192,22 @@ public class PacketHandler {
                     canPlace = c != cp.getColor() && c != 0xFF && c != -1;
                 }
                 int c_old = c;
-                /*if (!canPlace && (c == 0xFF || c == -1)) {
-                    // tendril expansion!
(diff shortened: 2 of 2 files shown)


Sorunome:
nonono the idea was to use the placemap and if the placemap on that index is e.g. 1 we use tendrils - that way we can have mixed canvases!

Am 4. September 2018 03:37:10 MESZ schrieb Andrew Black <notifications@github.com>:

FlyingSixtySix:
> if the placemap on that index is e.g. 1 we use tendrils
I don't understand. How can it be mixed? From what I see - either the entire canvas is tendril-only, tendril around an area only, or not. Also, what do you mean by _that index?_ Index of the pixel the user wants to place at?

Sorunome:
```
if (App.getHavePlacemap()) {
                    canPlace = c != cp.getColor() && App.getPlacemap(cp.getX(), cp.getY()) == 0;
                } else {
                    canPlace = c != cp.getColor() && c != 0xFF && c != -1;
                }
```

if App.getPlacemap(x, y) == 2 you do tendril expansion
that way we have 0 - normal place, 1 - no place, 2 - tendrils

Am 4. September 2018 07:32:49 MESZ schrieb Andrew Black <notifications@github.com>:

FlyingSixtySix:
`if (App.getConfig().getBoolean("tendrils") && App.getPlacemap(cp.getX(), cp.getY()) == 2) {`
So something like this instead?

Sorunome:
the config setting `tendrils` is obsolete with that.

Try storing the result if App.getPlacemap in a variable and do a switch for what to do on what behaviour type

Am 4. September 2018 07:47:24 MESZ schrieb Andrew Black <notifications@github.com>:

FlyingSixtySix:
Is what's in the latest commit what you were speaking off?

Sorunome:
yes exactly! if that works, that is xP

Am 4. September 2018 08:04:16 MESZ schrieb Andrew Black <notifications@github.com>:

FlyingSixtySix:
I'll do some testing in the morning. 👍

FlyingSixtySix:
I don't think this will work - `public static int App.getPlacemap(int x, int y)` returns a 0 or 1, based on the alpha value (cast to int).

Sorunome:
nope it can be anything from 0 - 255 https://github.com/xSke/Pxls/blob/master/src/main/java/space/pxls/App.java#L318-L320

PR #147, opened 2018-10-18
Add ability to lazy load iframes in drawers

TheRebelG0d:
This PR adds the ability to lazy load iFrames in drawers. This is in mainly so that the discord widget can be lazy loaded, as on page load we're requesting around 350+ profile pics.

This is achieved by adding an event to the drawers system that triggers when open state is toggled. iframes should then be reworked from
```html
<iframe src="#"></iframe>
```
to
```html
<iframe data-lazysrc="#"></iframe>
```


resources/public/info.html
```diff
@@ -49,3 +49,6 @@ <h1>Donate</h1>
 	</form>
     -->
 </div>
+<div>
+	<iframe data-lazysrc="https://discordapp.com/widget?id=191429241517244417&theme=dark" width="350" height="500" allowtransparency="true" frameborder="0"></iframe>
```
FlyingSixtySix: Remove this line.

PR #157, opened 2018-11-15
Add rule dropdown to ban menu

FlyingSixtySix:
This commit adds the features from #155 to the ban prompt.
Other than that, it adds a missing semicolon.


FlyingSixtySix:
I committed and merged before I was ready; functionality is restored with #158.

PR #166, opened 2018-11-18
Add client-side virginmap view

FlyingSixtySix:
This commit adds a heatmap-like virginmap view. The green indicates virgins, and black indicates non-virgins.


FlyingSixtySix:
While I was working on this, I noticed that clearing the heatmap (via hotkey, with the option enabled) throws this error (in the `heatmap` `_clear()` function):

screen shot 2018-11-18 at 2 30 53 pm

PR #188, opened 2019-03-19
Logging system tweaks

GlowingSocc:
**This change is not retroactive and should not be merged until a canvas reset.**

This PR brings the following:
* Move app logging to its own log handler
* Convert pixel logger to TSV
* Always attach an action to the pixel logger
* Log shadowbanned pixels in their own file for monitoring

These are basically just QoL updates for our current logging system in terms of parsing and data displayed.
This is achieved by updating pixel logging into a tab-separated format, as well as ensuring there's always an `action` column attached.

This PR also removes `clearing old sessions` from the pixel log, and moves most app logging to an actual logger, as well as logs shadowbanned pixels in their own file for monitoring.

Example new log structure:
```
DATE	USER_NAME	X	Y	COLOR	IP	ACTION


2019-03-19 02:14:30,239	<server>	0	0	4		console nuke
2019-03-19 02:14:40,305	USER	0	0	5	192.168.1.1	user place
2019-03-19 02:14:53,760	USER	1	1	5	192.168.1.1	mod overwrite
2019-03-19 02:15:56,043	USER	164	217	5		rollback
2019-03-19 02:16:06,081	USER	164	217	4		rollback undo
```
`console nuke` is the `nuke` command from console
`user place` is a normal user pixel place
`mod overwrite` is a mod using the 'Override CD' setting
`rollback` is a ban rollback (removed placed pixels on permaban/etc)
`rollback undo` is an unban rollback (restored pixels after unban).

The difference being IPs are no longer empty on undo, and actions (undo/nuke), etc are always displayed in the rightmost column. This makes parsing for staff reasons much cleaner, and general consumption much easier.

### DIFF
--- a/resources/log4j2.xml
+++ b/resources/log4j2.xml
@@ -6,7 +6,19 @@
 
         <File name="Pixels" fileName="logs/pixels.log">
             <PatternLayout>
-                <Pattern>%d %m%n</Pattern>
+                <Pattern>%d&#9;%m%n</Pattern>
+            </PatternLayout>
+        </File>
+
+        <File name="ShadowbannedPixels" fileName="logs/shadowbannedPixels.log">
+            <PatternLayout>
+                <Pattern>%d&#9;%m%n</Pattern>
+            </PatternLayout>
+        </File>
+
+        <File name="App" fileName="logs/app.log">
+            <PatternLayout>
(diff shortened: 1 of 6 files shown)


Undoundoundo:
missing ACTION: 'user undo', which is a user hitting undo after placing a pixel, which effectively reverts that pixel to its previous state.

PR #209, opened 2019-08-31
Added ability to set a Discord name (name#discrim) for pixel and lookups

netux:
~~So I finally decided to install IntelliJ to do this~~

- This adds a new field to the users table: `discord_name VARCHAR(37)`
- User is forced to use "{name}#{discriminator}" syntax
- Name can be updated in settings tab, under "Account settings"
- Name appears on pixel lookups as well as admin lookups

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -2285,9 +2285,9 @@ window.App = (function () {
                     // Register default hooks
                     self.registerHook(
                         {
-														id: "coords",
-														name: "Coords",
-														get: data => $("<a>").text("(" + data.x + ", " + data.y + ")").attr("href", coords.getLinkToCoords(data.x, data.y)),
+                            id: "coords",
+                            name: "Coords",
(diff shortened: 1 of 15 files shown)


resources/public/pxls.js
```diff
@@ -2206,6 +2206,10 @@ window.App = (function () {
                         if (data) {
                             return $.map(self.hooks, function (hook) {
                                 const get = hook.get(data);
+                                if (!get) {
```
GlowingSocc: If the hook result returns falsey then this short circuits, so e.g. if `pixel_count` is `0` then it doesn't show up in the shift+click result. This can also impact custom hooks that weren't expecting this behavior, e.g. "is on watchlist: false". This would fix most cases:
```suggestion
                                if (get == null) {
```

netux: Even better, have another function return whenever we should show it or not.

I might get to doing that tonight. For now though, I'll apply this.

PR #245, opened 2019-12-11
Attempt to Workaround for Chrome-based browsers offsetting the canvas by a pixel

netux:
- Increase or decrease #board-container size depending on window and board size
- Ability to toggle workaround on and off

This is probably missing browsers like Safari and Opera, which might be affected by the bug but are discriminated out of the workaround for now.

Please test to see if this works for all canvas and window sizes (tested with odd and even of both).

### DIFF
--- a/resources/public/index.html
+++ b/resources/public/index.html
@@ -288,12 +288,14 @@
                         </div>
                     </div>
 
-                    <label class="vertical-spacing"><input type="checkbox" id="color-brightness-toggle"> Enable Color Brightness</label>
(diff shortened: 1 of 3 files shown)


FlyingSixtySix:
Pixels are offset to the right by one (Chrome 79.0.3945.79) with an `innerWidth`/`innerHeight` of 1159/443 respectively. They're also shifted one down and to the right with 955/374. Cannot reproduce on Firefox 71.0 or iOS 13.2.3 Safari.

FlyingSixtySix:
Fix tested and working on latest Chrome on macOS.

PR #261, opened 2020-01-05
Add domain white/blacklist config options

FlyingSixtySix:
This commit adds the ability to whitelist and blacklist email domains from signing in/up.


FlyingSixtySix:
Closed; untested + irrelevant to production.

PR #276, opened 2020-01-27
Gulp Frontend

FlyingSixtySix:
This PR removes `minify-maven-plugin` in favor of `frontend-maven-plugin`, handing off the frontend build process to Gulp. The new plugin will install Node.JS v12.14.1 (LTS as of 2019-01-23), install dependencies, and run `gulp build`. The gulp build configuration copies HTML files, sourcemaps and minifies CSS files, and copies + minifies JS files. The steps for ESLinting are commented out while production remains completely uncompliant with the new standard.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "pxls",
+  "version": "1.0.0",
+  "dependencies": {
+    "eslint": "^6.8.0",
+    "eslint-config-semistandard": "^15.0.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "pxls",
+  "version": "1.0.0",
```
netux: It's always good practice to add `"private": true` on this file

PR #282, opened 2020-02-14
Moderator Chat Lookup

GlowingSocc:
A new action has been added to the popup menu for Staff that allows them
to pull a user's last 100 chat messages as well as their chat bans. This
should help cut back on the time spent lookup up things such as 3-star
messages.

### DIFF
--- a/src/main/java/space/pxls/App.java
+++ b/src/main/java/space/pxls/App.java
@@ -1,7 +1,6 @@
 package space.pxls;
 
 import com.google.gson.Gson;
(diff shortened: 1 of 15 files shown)


src/main/java/space/pxls/App.java
```diff
@@ -1,6 +1,7 @@
 package space.pxls;
 
 import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
```
netux: Unused import

PR #290, opened 2020-02-27
Pop a small alert for staff on new received report

GlowingSocc:
This PR introduces a set of UI elements and functions which give us the ability to pop sliding banners with optional auto-hide timeouts. They are currently only hooked up to show an alert when a canvas and/or chat report is received by the server to staff, however they will be used in the future to replace most "job done" modals. That is outside the scope of this PR though, and this PR is considered a blocking dependency for those.

### DIFF
--- a/resources/public/SLIDEIN.css
+++ b/resources/public/SLIDEIN.css
@@ -1,0 +1,90 @@
+@keyframes slidein-slide {
+    0% {
+        transform: translateY(-2em);
+    }
+    100% {
+        transform: translateY(0em);
+    }
+}
+
+.slidein {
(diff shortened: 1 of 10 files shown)


GlowingSocc:
Slideins have been moved to their own file for future use, no functional changes have been made. I do not believe a re-review is necessary.

netux:
If the SlideIn code is put on another file, why not put the styles on another file too?

GlowingSocc:
> If the SlideIn code is put on another file, why not put the styles on another file too?

I guess a re-review would have been good, forgot I included the CSS in index.css. another commit incoming

PR #291, opened 2020-02-27
Tweak modal close button styles

GlowingSocc:
The modal close button was escaping its parent's boundaries and causing
issues for mobile users. The new close button now squishes into the top
right corner so that it is always accessible and doesn't overflow.

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -6201,6 +6201,7 @@ window.App = (function () {
                         escapeClose: true,
                         clickClose: true,
                         showClose: true,
+                        closeText: '<i class="fas fa-times"></i>',
                     }, {removeOnClose: true}, opts);
                     if (!document.body.contains(modal)) {
                         document.body.appendChild(modal);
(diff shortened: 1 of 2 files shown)


resources/public/pxls.js
```diff
@@ -6201,6 +6201,7 @@ window.App = (function () {
                         escapeClose: true,
                         clickClose: true,
                         showClose: true,
+                        closeText: '<i class="fas fa-times"></i>',
```
netux: Unused var.

netux: My bad 🤦‍♂

GlowingSocc: For future reference, `closeText` is used upstream by the modal library. Since they inject the string directly into tag creation I'm replacing the text with an icon.
Ref: https://github.com/kylefox/jquery-modal/blob/a8b9c9b01961c0a3c6676fcc8fe6cc5b8c74cb5c/jquery.modal.js#L142

PR #296, opened 2020-03-02
Profiles + Factions

GlowingSocc:
**Warning: Requires DB Maintenance.**
*This is not a backwards compatible update.*

This PR brings the following:
* User Profiles
    * User profiles are pretty bare bones at this point in time, but will be expanded on in the future.
* A faction system
    * Faction management is done from user profiles
    * Factions display in chat, on lookups, and on profiles

Factions require a new field on the `users` table. This field references the `faction` table, so pxls should be dry-run before updating the database.
```sql
ALTER TABLE users ADD displayed_faction INT;
```


resources/cache.ccf
```diff
@@ -0,0 +1,29 @@
...
+jcs.region.factions.cacheattributes.MaxSpoolPerRun=500
+
+jcs.region.users=
+jcs.region.users.cacheattributes=org.apache.commons.jcs.engine.CompositeCacheAttributes
+jcs.region.users.cacheattributes.MaxObjects=250
+jcs.region.users.cacheattributes.MemoryCacheName=org.apache.commons.jcs.engine.memory.lru.LRUMemoryCache
+jcs.region.users.cacheattributes.UseMemoryShrinker=true
+jcs.region.users.cacheattributes.MaxMemoryIdleTimeSeconds=3600
+jcs.region.users.cacheattributes.ShrinkerIntervalSeconds=60
+jcs.region.users.cacheattributes.MaxSpoolPerRun=500
```
GlowingSocc: This region is added for future use but not yet hooked up. Refactoring the user manager is not a priority with this PR unless we want to lump the two together.

resources/public/profile/js/views/profile.js
```diff
@@ -0,0 +1,406 @@
...
+                  });
+                  btnJoin.addEventListener('click', async function() {
+                    if (await popConfirmDialog('Are you sure you want to join this faction?') === true) {
+                      let res;
+                      try {
+                        res = await putJSON(`/factions/${x.id}`, {joinState: true});
+                      } catch (e) {}
+                      if (res && res.success === true) {
+                        alert('Faction joined. The page will now reload.');
+                        document.location.href = document.location.href; //TODO replace with slidein
```
GlowingSocc: The slideins mentioned are being included in #290, see a01f45a for the file inclusion.

GlowingSocc:
**Force Push**: Rebase master, resolve `reference.conf` conflicts.

src/main/java/space/pxls/server/packets/chat/ChatMessage.java
```diff
@@ -1,24 +1,30 @@
...
     public List<Badge> badges;
     public List<String> authorNameClass;
     public Number authorNameColor;
+    public StrippedFaction strippedFaction;
 
-    public ChatMessage(String nonce, String author, Long date, String message_raw, List<Badge> badges, List<String> authorNameClass, Number authorNameColor) {
+    public ChatMessage(String nonce, String author, String tag, Long date, String message_raw, List<Badge> badges, List<String> authorNameClass, Number authorNameColor, Faction faction) {
         this.nonce = nonce;
         this.author = author;
+        this.tag = tag;
```
netux: Could get the tag from `strippedFaction` instead.

GlowingSocc: Apparently I'm not actually using `.tag` anywhere, and everything is sourced from `strippedFaction`. Not sure how I missed that. I've removed the tag property entirely.

resources/public/profile/js/views/root.js
```diff
@@ -0,0 +1,199 @@
...
+
+async function postJSON(url, data = {}) {
+  return _doPostLike(url, 'POST', data);
+}
+
+async function patchJSON(url, data = {}) {
+  return _doPostLike(url, 'PATCH', data);
+}
+
+async function _doGetLike(url, type) {
```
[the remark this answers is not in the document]
GlowingSocc: Pulled from another one of my projects, and xhr does the job well enough imo. The `fetch` interface used to have problems with cookies, not sure if that's still the case, but we're same-domain either way. I'm just more comfortable with XHR overall.

GlowingSocc:
Pushed some commits to address comments. Ready for re-review.

PR #318, opened 2020-04-28
Implement CSS Variables

BlankParenthesis:
This implements CSS variables into both the main stylesheet and themes.
[https://caniuse.com/#feat=css-variables](https://caniuse.com/#feat=css-variables)
If CSS variables are deemed widely supported enough, the backup definitions in style.css and admin.css can be safely removed. With these backup definitions, browsers without support for CSS variables will be able to use the default theme without issue but will not be able to use other themes.

### DIFF
--- a/resources/public/style.css
+++ b/resources/public/style.css
@@ -809,7 +809,7 @@ body:not(.show-placeable-bubble) #placeableInfoBubble {
     text-decoration: underline;
 }
 
-input[type="text"], input[type="number"], textarea, select, button {
+input[type="text"], input[type="number"], textarea, select {
     padding: 0.5em;
     border: 1px solid #ccc;
     border: 1px solid var(--input-border-color);
--- a/resources/public/style.css
+++ b/resources/public/style.css
@@ -743,6 +743,7 @@ body:not(.show-placeable-bubble) #placeableInfoBubble {
     line-height: 24px;
     border-radius: 16px;
     text-align: center;
+    color: #000;
     background-color: white;
     border: 1px solid black;
     padding: 0 8px;
--- a/resources/public/style.css
+++ b/resources/public/style.css
@@ -120,7 +120,6 @@ html {
     --pane-button-background: linear-gradient(0deg, #eee, #ccc, #eee);
     --pane-button-background-hover: linear-gradient(0deg, #ddd, #bbb, #ddd);
     
-    /* [  ]: These should be implemented in some standard way rather than as an exception imo */
     --notification-background-color: #f5f5f5;
     --notification-border-color: #8f8f8f;
     --notification-title-text-color: inherit;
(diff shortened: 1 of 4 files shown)


resources/public/style.css
```diff
@@ -1325,21 +1537,27 @@ li.chat-line .actions [data-action]:not(:last-of-type):not(:only-child) {
...
 }
 
 #typeahead ul li.active {
     background-color: #828282;
+    background-color: var(--chat-typeahead-item-background-color-active);
 }
 
 /* PINGS */
 
+/* NOTE: this seems to not respond to css vars (at least on firefox when I tested it) so I'm leaving it fixed for now */
```
netux, line 1560: From what I've tested browsers act differently depending on whenever the variables are declared before or after the animation 👍

resources/public/style.css
```diff
@@ -1475,8 +1698,16 @@ ul.chat-history, ul.chatban-history {
     text-shadow: none !important;
 }
 
-.emoji-picker {
+/* preselecting body makes this a more specific selector than the built-in one which would normally override this since it's inline */
+/* the same principle is why each of the following preselect .emoji-picker */
```
netux, line 1702: 👍

resources/public/style.css
```diff
@@ -1598,6 +1907,7 @@ ul.chat-history, ul.chatban-history {
     user-select: none;
 }
 
+/* NOTE: popup-actions seems unused */
```
netux, line 1910: 👍 they were used before Socc reworked popups. Can be removed.

resources/public/themes/dark.css
```diff
@@ -1,156 +1,107 @@
...
+    --modal-background-color: #3d3d3d;
+    --modal-headfoot-background-color: #292929;
 }
 
 hr {
     border: none;
     border-top: 1px solid #000;
 }
 
+/* probably supposed to be .popup-timestamp-header (--popup-timestamp-text-color) */
```
netux, line 100: 👍 not used, can be removed from here and style.css as well.

resources/public/style.css
```diff
@@ -618,22 +800,27 @@ body:not(.show-placeable-bubble) #placeableInfoBubble {
...
+    color: var(--copypulse-color);
 }
 
 .userinfo .logout {
     font-size: 0.75em;
     text-decoration: underline;
 }
 
-input[type="text"] {
+input[type="text"], input[type="number"], textarea, select, button {
```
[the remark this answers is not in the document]
BlankParenthesis: Not all `button` elements are not styled with `--button-*` variables. Pxls uses seems to have three styles of button currently - `button`s with the button class (which would work as you describe), non-button elements with the button class, and `button`s without the button class.

I would seek to resolve this issue with a merging of the uses so that every button just uses the `button` element and nothing more. This is not something I wanted to accomplish with this PR however.

For the moment, I decided to style buttons with other input elements since that would more preserve how they currently appear.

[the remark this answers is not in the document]
netux: After some internal talk we have decided to use the `.button` class to denote buttons so that we can display non-`<button>` elements as a button, on a bootstrapy-way.

Currently `<button>`s without the `.button` class display as the default browser button. So this change wouldn't preserve their current look, it'd change it completely.
One of my major complaints here is how the buttons don't have any hover feedback. Browser buttons and buttons with the `.button` class light up in some way or the other, indicating they are clickable.

So far I've identified two `<button>`s that don't have a `.button` class on them: `#hvmapClear` and `#template-width-reset`. I plan on adding a `.button` class to them on a later PR and having them in the default browser style would also help me remember to do so every time I see them.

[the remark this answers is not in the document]
BlankParenthesis: I disagree with your conclusions about the semantics but I can remove the button selector from this line if you insist.

netux:
Looks good! One last thing: could you remove all those dev comments, or prepend them with `NOTE([ ]): `?

PR #321, opened 2020-05-02
Add faction tag unicode ranges

BlankParenthesis:
This leans more on the conservative side but does also allow some full blocks that may not be needed such as the mathematical operators block.

We might very well desire to add more entire blocks like the entirety of currency rather than just Euros and Bitcoin.

This also neglects to support various languages that should probably be supported in favour of focusing more on symbols and punctuation that may be missing.

### DIFF
--- a/src/main/java/space/pxls/user/Faction.java
+++ b/src/main/java/space/pxls/user/Faction.java
@@ -243,13 +243,8 @@ public class Faction {
         new int[] {0x2190, 0x2199}, // simple arrows
         new int[] {0x231A, 0x231B}, // watch, hourglass from misc technical
         new int[] {0x23EA, 0x23FA}, // more emoji from misc technical
-        new int[] {0x2460, 0x2468}, // circled numbers up to 9
-        new int[] {0x24B6, 0x24EA}, // circled alphabet + circled zero
         new int[] {0x2600, 0x27BF}, // misc symbols (♥), dingbats (sparkle, heavy heart)
         new int[] {0xFE00, 0xFE0F}, // variation selectors (heart color)
-        new int[] {0xFF01, 0xFF5E}, // fullwidth latin characters
-        new int[] {0xFFE1}, // fullwidth pound
-        new int[] {0xFFE5, 0xFFE6}, // fullwidth yen + won
         new int[] {0x1F000, 0x1FAFF} // emoji
     );
     private static boolean checkCodepoints(String input) {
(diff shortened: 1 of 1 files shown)


netux:
Oh wow, this is a lot more than I was expecting.
A few points:
* Quite a few of these render as very wide characters, which can lead to spammy tags like "ＮＥＴＵＸ" (full-width range) or "⏠⏡⏠⏡⏠" (from the misc range).
* The full-width and circled alphabet could also be used to bypass the text filters we have in place.
* All the musical notes (also the musical cleff), and these characters from the misc technical range `⎷ ⎸ ⎹ ⎺ ⎻ ⎼ ⎽` don't render correctly on the devices I've tried.

Overall, I'm not sure how much use most of the math symbols and misc technical symbols will get. The quotation marks might be an overkill as well.

BlankParenthesis:
In regards to the full-width characters, I expect that we should eventually either support all languages anyway or make this configurable at the very least. One of these languages is Japanese which makes use of full-width characters.

I think the correct choice is not to prevent everyone from using them but to moderate it appropriately as you would the rest of chat. This also applies to the filters, speaking of which - do the current filters consider Latin extended characters for this purpose?

I understand the concern about musical notes and there are some appropriate emoji that can be used as a substitute. As for the technical symbols, there are some characters in that range usually considered emoji: 
⌚⌛⏰⏱⏲⏳ and ⏩⏪⏫⏬⏭⏮⏯⏸⏹⏺ that should probably be included at least.

Let me know what you think and we can adjust the ranges appropriately.

netux:
> I expect that we should eventually either support all languages anyway
From what I've gathered, support for multiple languages is something we don't really want, as it could make moderation difficult.
In fact, part of why there are restrictions here is to make moderation easier.

> [...] or make this configurable at the very least
That'd actually be swell, so we don't have to make PRs everytime we change the codepoints.

> do the current filters consider Latin extended characters for this purpose?
The filter are configurable so they can be made into accepting those; but so far it hasn't been necessary.

> As for the technical symbols, there are some characters in that range usually considered emoji
Yes, I saw them. Those can stay; but the rest of the misc seemed very useless.

BlankParenthesis:
My primary concern for supporting multiple languages is that this is supposed to be an open project that anyone can use (and people have used it for their own purpose)

It seems like you prefer the configuration way of solving this but there's something to be said for not needing to configure a potentially difficult option and have a good built-in setting.

netux:
Sorry for the late response:
While yes, we can change the filter regex's to suit our needs, I failed to mention adding the new alphabets adds complexity to the regex's. In fact, I just tried doing one adding the new symbols to one of them and it became an unreadable mess.

Besides, the more I talk with staff the more they say they wouldn't prefer them for now as they seem unnecessary and a bit spammy.

Since the next canvas reset is so soon I'd suggest removing both circled and full-width alphabet ranges for now and working on their addition in another PR that can solve the complexity of the filters and/or make these ranges configurable.

BlankParenthesis:
I did think that's what you meant - I was sort of intending to poke at the fact that you didn't really have a proper system but that can be addressed elsewhere.

I'll remove them for now so we can merge this and look at solving the other issue brought here another time.

PR #322, opened 2020-05-02
Blue theme

cp-codes2:
added blue theme to the theme section and added the css file.

 It _should_  be the same as my last pr. someone will also have to see the admin and mod ui's to make sure they look good with this.

### DIFF
--- a/resources/public/admin/admin.css
+++ b/resources/public/admin/admin.css
@@ -27,7 +27,7 @@
     visibility: visible;
 }
 
-.admin > h1 {
+.admin > h2 {
     display: block;
     font-size: 20px;
     font-weight: 700;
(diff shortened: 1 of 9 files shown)


FlyingSixtySix:
Thanks for your submission! I found a few issues I think should be addressed:

1. Button text (`--button-text-color`) does not contrast enough with the button background:

This affects all button elements, including the moderator information panel:


2. The moderator side-panel (background: `--admin-background`, text: `--admin-text-color`) are still default:


3. The odd alternating message color (`--chat-odd-background-color`), the separator color (`--chat-separator-color`), the server action text color (`--chat-server-action-text-color`), and the ping highlight color (`--ping-highlight-color`) look out of place being inherited from dark theme:


4. The emoji picker selection text color (`--emoji-picker-section-text-color`) does not contrast well with the box - maybe it would look better darker?

cp-codes2:
like I said in the description of my most recent commit (a few minutes ago) it looks like  --button-text-color, --admin-background, and --admin-text-color never existed in the first place unless they were controlled by something else that I have overlooked. Im guessing that's because the other themes used the default style for those parts.

PR #324, opened 2020-05-04
CSS cleanup

BlankParenthesis:
This touches basically everything with the goal of reducing the amount of unique code used for UI components.

At the same time, I couldn't help myself from cleaning up some somewhat unrelated things so there's some miscellaneous changes and fixes thrown in because that's *definitely how this is supposed to work 🙃*.

The overall gist of the reduction in UI components is that before there were 8 identifiable components: 
- Side Panels
- UI Bubbles
- Panel Articles/Sections
- Lookup Dialog
- System messages
- Modal Dialogs
- Notifications
- Popups

With special cases arising for moderation tools such as user lookups, the mod sidepanel, and chatban history.

This PR reduces this to just 4 core components:
- UI Bubbles (which now includes the mod sidepanel)
- Panels (side panels, modal dialogs, and popups)
- Panel Articles/Sections (which are used by notifications and chatban history)
- Floating Panels (lookups and system messages)

Perhaps the most significant other changes are those made to buttons. The summary of this is that anything a user is expected to click on to perform an action is now a button. This is mostly just an HTML semantics things which isn't even all that related to CSS but resolving it at the same time resulted in more consistent button styling methods for traditional text-buttons.

Hopefully this can be merged without too much change, but I wouldn't be surprised to be told "go back and break it into more appropriate pieces".

### DIFF
--- a/resources/public/style.css
+++ b/resources/public/style.css
@@ -501,11 +501,11 @@ body .message {
 
 #ui {
     position: fixed;
-    width: 100%;
-    height: 100%;
     top: 0;
+    bottom: 0;
     left: 0;
-    transition: width ease-in-out .25s, left ease-in-out .25s;
+    right: 0;
+    transition: left ease-in-out .25s, right ease-in-out .25s;
     user-select: none;
     -moz-user-select: none;
     -ms-user-select: none;
@@ -1037,7 +1037,10 @@ footer {
 aside.panel {
     position: fixed;
     top: 0;
-    max-width: 30%;
+    /* NOTE ([  ]): Ideally this would be max-width and we could set an overriding fixed min-width.
+                    This is just width because the main ui currently can't know how much of the min-with is used
+                    and can't adjust its own width accordingly. */
+    width: 30%;
     height: 100%;
     z-index: 21;
     cursor: initial !important;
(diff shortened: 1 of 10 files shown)


FlyingSixtySix:
I think `.floating-panel` should be a bit more compact:

- Reduce the value of `margin-top` in `.floating-panel .content > div:not(:first-of-type)` to `5px`
- Change `padding` in `body .message` to `padding-bottom: 10px;` (If you feel it should be explicit, `padding: 0 0 10px;`)
- Remove `min-width` in `body .message`
- Remove `margin-bottom` in `.floating-panel .content`

resources/public/style.css
```diff
@@ -1753,8 +1741,8 @@ body .emoji-picker {
 .emoji-picker .emoji-picker__search-container input {
     border: 1px solid #CCCCCC;
     border: 1px solid var(--emoji-picker-search-input-border-color);
```
FlyingSixtySix, line 1741: `--emoji-picker-search-input-border-color` does not exist.

BlankParenthesis:
I've done mostly as requested but kept some of the bottom margin in `.floating-panel .content` since this padding is still somewhat important. This is most noticeable in lookups where it originates from:
image
(Image with no margin)

PR #327, opened 2020-05-06
Fix various things being missing from mod lookup

BlankParenthesis:
This fixes two minor bugs in the mod lookup popup:
- The input for sending a user an alert was previously missing a type in both the pixel lookup and the mod lookup. 
- The login info for a user in the mod lookup was missing.

### DIFF
--- a/resources/public/admin/admin.js
+++ b/resources/public/admin/admin.js
@@ -11,7 +11,7 @@
   };
   const sendAlert = function(username) {
     if (admin.user.getRole() === 'TRIALMOD') return '';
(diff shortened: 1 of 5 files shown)


FlyingSixtySix:
The parameter `String method` of `ServerUserInfo` and `ServerUserPixelInfo` was meant to be displayed under `login`, but I haven't looked into why that hasn't been appearing.

BlankParenthesis:
Ah, I see that.

I'm not sure why you wouldn't display the info in the same way as you do in pixel lookups but my guess for why it doesn't show up is that it would be serialised as `method` but the client looks for `data.login`. Upon quickly testing, this is the case and can also be easily fixed.

PR #337, opened 2020-05-12
Refactor settings

BlankParenthesis:
This adds a new component-level object `settings` which makes managing and keeping track of various settings and controls for them much easier.

Before the process for creating a setting was thusly:
```js
let settingValue = ls.get('setting-key');
if(settingValue == null) {
  // or some other invalid value
  settingValue = defaultValue;
}
$('#settingControl').val(settingValue);
$('#settingControl').change(function() {
  ls.set('setting-key', this.value);
  // do onchange things
});
```

To do something similar with this PR, we would do the following:
```js
settings.component.feature.subfeature.setting = setting('setting-key', SettingType.NUMBER, defaultValue, $('settingControl'));
settings.component.feature.subfeature.setting.listen(function() {
  // do onchange things
});
```

One major benefit of this is the programmatic changing of settings. It is now possible to in one call enable the heatmap, record the setting to localstorage, and update the checkbox for the heatmap being enabled. This works to same for all other settings.

```js
settings.board.heatmap.enable.set(true);
```

In regards to this PR, perhaps my most significant doubts are in the naming of settings keys. Since these names have changed there is code to convert over old ones such that users settings should be retained, but the new choice of keys is still up for debate. As demonstrated above, I have decided to go with `component.feature.subfeature.setting`. 

Even if we agree on this naming though, the decision of which component a given setting counts for is not clear. For example, audio is its own component but many settings that would be associated with a different component may be audio related. To be more specific, a key such as `chat.pings.audio.enable` could just as easily be `audio.pings.enable`—and perhaps should be, given audio is more the relevant category.

One final thing to mention is that this also flips some of the settings from being check-to-disable to being more consistent by being check-to-enable. An example of this is audio becoming "Enable audio" rather than "Mute audio".

Regardless, this code should make handling of settings much easier and more consistent going forward.

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -539,7 +539,7 @@ window.App = (function() {
     };
 
     // these are the settings which have gone from being toggle-off to toggle-on
-    const flippedmappings = ['audio_muted', 'increased_zoom', 'autoReset'];
+    const flippedmappings = ['audio_muted', 'increased_zoom', 'autoReset', 'canvas.unlocked'];
 
     // Convert old settings keys to new keys.
     Object.entries(keymappings).forEach((entry) => {
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -543,9 +543,7 @@ window.App = (function() {
 
     // Convert old settings keys to new keys.
     Object.entries(keymappings).forEach((entry) => {
-      // This is the best we can do to check existance without reacing into localStorage ourselves.
-      // If there we were a ls.has function, that should be used here.
-      if (ls.get(entry[0]) !== null) {
+      if (ls.has(entry[0])) {
         const oldvalue = ls.get(entry[0]);
         ls.set(entry[1], flippedmappings.indexOf(entry[0]) === -1 ? oldvalue : !oldvalue);
         ls.remove(entry[0]);
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -359,17 +359,17 @@ window.App = (function() {
     const filterInput = function(controls, type) {
       switch (type) {
         case Type.TOGGLE:
-          return $('input[type=checkbox]').filter(controls);
+          return controls.filter('input[type=checkbox]');
         case Type.RANGE:
-          return $('input[type=range]').filter(controls);
+          return controls.filter('input[type=range]');
         case Type.TEXT:
-          return $('input[type=text]').filter(controls);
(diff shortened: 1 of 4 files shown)


resources/public/pxls.js
```diff
@@ -313,6 +313,394 @@ window.App = (function() {
...
+            toAdd.on(changeEvents, changeFunction);
+
+            if (disconnectOnRemove) {
+              // We don't want to leak memory if the code that adds controls is so inconsiderate as to not clean up after itself.
+              // So here we listen for changes in the DOM and remove them from the controls list if they are no longer attached.
+              // JQuery UI has a remove event that we can listen for if we have JQuery UI - that would make this cleaner.
+              toAdd.each((_, c) => {
+                const observer = new MutationObserver(function() {
+                  // NOTE ([  ]): Modal seems to fire many add/remove mutations for the panel itself when adding a new one.
+                  //              This means we can't just listen for remove mutations and filter accordingly.
```
netux, line 465: I'm not sure about this whole MutationObserver section.

Having a MutationObserver that listens to all DOM adds/removes for each setting we have (and in some cases multiple listeners for a setting) sounds heavy on resources.

I think a better solution to this would be to have whatever is adding these `controls` also remove them.
I think this was implementing for chat settings, so having the module that handles them listen for when the Chat Settings popup closes and remove all elements manually would work much better.

resources/public/pxls.js
```diff
@@ -313,6 +313,394 @@ window.App = (function() {
...
+          changeEvents = 'input change';
+          break;
+        case SettingType.TOGGLE:
+          changeFunction = (evt) => self.set(evt.target.checked);
+          break;
+      }
+
+      const self = {
+        get: get,
+        set: set,
```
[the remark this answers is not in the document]
BlankParenthesis: I decided not do this on purpose. This was to try and keep a consistent style with the rest of the code.

resources/public/pxls.js
BlankParenthesis: I believe the reason I decided not to do this was mostly that there was no reliable way to know when the settings popup was done - unless you suggest modal tell settings about its state or that modal track this itself.

BlankParenthesis: I've implemented this as best I can for now.

Another option is just having the controls leak until the next time `set` is called at which point all controls are validated to be in the DOM and any that aren't are removed.

PR #342, opened 2020-05-15
Fix chrome enabling multiple themes at once

BlankParenthesis:
Since Chrome does not fire the load event for elements that have already loaded being added back to the DOM, the code that cleaned up previous themes would not fire when switching back to a theme that had previously been loaded.

This fix changes the way themes are loaded such that they stay on the page and are disabled with the `disabled` property until needed again. In addition, a `loaded` variable is also stored to determine if the app needs to add the newly selected theme to the DOM.

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -3075,6 +3075,7 @@ window.App = (function() {
       _initThemes: function() {
         for (let i = 0; i < self.themes.length; i++) {
           self.themes[i].element = $('<link data-theme="' + i + '" rel="stylesheet" href="' + self.themes[i].location + '">');
+          self.themes[i].loaded = false;
           self.elements.themeSelect.append($('<option>', {
             value: i,
             text: self.themes[i].name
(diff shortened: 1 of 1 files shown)


resources/public/pxls.js
```diff
@@ -3361,6 +3346,45 @@ window.App = (function() {
...
+          return;
+        }
+        if (!(index in self.themes)) {
+          return console.warn(`Tried to load invalid theme "${index}"`);
+        }
+        const theme = self.themes[index];
+        if (theme.loaded) {
+          self.enableTheme(index);
+        } else {
+          await new Promise((resolve, reject) => {
```
netux, line 3363: Why does this need to be wrapped inside a Promise?

BlankParenthesis: It doesn't strictly have to be, but this way functions calling `loadTheme`  can await until the theme actually gets loaded.

Unless `JQuery.one` returns a Promise which don't see any documentation of.

netux: If some code ever needs to wait for a theme to load, then this can be readded; but right now it has no use.

BlankParenthesis: Can I get a second opinion on this?
Waiting for the theme to load before returning seems like the *correct* behaviour for this function regardless of whether any code makes use of that feature.

FlyingSixtySix: I think it'd be good to have anyway for the reason @BlankParenthesis noted.

PR #347, opened 2020-05-20
Improve support for multiple tabs open

netux:
* Use Service Workers to detect which tab has focus when possible
* Expose `App.uiHelper.tabId` and `App.uiHelper.tabHasFocus()`
* Clear chat notifications in other tabs once the last messages are seen.


BlankParenthesis:
Since only one serviceworker can be registered for a given scope, it might be worth renaming the `focusworker.js` script to a more generic serviceworker name in case more uses for serviceworkers are desired in future.

PR #359, opened 2020-05-31
Adjust scale boundaries to [0.01, 50] everywhere posible

netux:
This PR:
* Adjusts the zoom lower bound to 0.01, so that values so small they seem null aren't possible.
* Replaces all direct assignments to `board.scale` with `board.setScale`s, so that the [0.01, 50] limit is enforced.

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -1671,7 +1671,7 @@ window.App = (function() {
       },
       setScale: function(scale, doUpdate = true) {
         if (settings.board.zoom.limit.enable.get() !== false && scale > 50) scale = 50;
-        else if (scale <= 0.01) scale = 0.01; // enforce the [0.5, 50] limit without blindly resetting to 0.5 when the user was trying to zoom in farther than 50x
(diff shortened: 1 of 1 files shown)


resources/public/pxls.js
```diff
@@ -1668,23 +1669,21 @@ window.App = (function() {
       getScale: function() {
         return Math.abs(self.scale);
       },
-      setScale: function(scale) {
+      setScale: function(scale, doUpdate = true) {
         if (settings.board.zoom.limit.enable.get() !== false && scale > 50) scale = 50;
-        else if (scale <= 0) scale = 0.5; // enforce the [0.5, 50] limit without blindly resetting to 0.5 when the user was trying to zoom in farther than 50x
+        else if (scale <= 0.01) scale = 0.01; // enforce the [0.5, 50] limit without blindly resetting to 0.5 when the user was trying to zoom in farther than 50x
```
BlankParenthesis, line 1674: The comment here should probably read `enforce the [0.01, 50] limit... ...to 0.01 when...`

PR #361, opened 2020-05-31
fixing blue theme

cp-codes2:
made everything darker and fixed: ping menu, ping color, emoji picker as well


cp-codes2:
looks like this now
Screen Shot 2020-05-31 at 3 27 33 PM

PR #363, opened 2020-06-01
Fix palette number changing with button font size

BlankParenthesis:
Most user agent buttons have their own size. This sets the palette number based on the root document text size instead of their parent element.

### DIFF
--- a/resources/public/style.css
+++ b/resources/public/style.css
@@ -712,7 +712,7 @@ body .message {
     color: #000;
     background-color: #fff;
     border: 1px solid #000;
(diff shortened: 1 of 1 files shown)


resources/public/style.css
```diff
@@ -712,8 +712,8 @@ body .message {
     color: #000;
     background-color: #fff;
     border: 1px solid #000;
-    width: 2em;
-    font-size: .75em;
+    width: 2rem;
```
netux, line 715: I'd reduce the width to about `1.5rem` as well

PR #368, opened 2020-06-02
Improve semantic structure of settings inputs

BlankParenthesis:
This does several things to improve the semantics and structure of input fields used for settings as well as improving consistency.

- Input elements are now wrapped in a label.
- Labels containing an input element have their text content wrapped by an element with the `label-text` class.
- Setting fields are now grouped using paragraph elements.
- Extra text that provide additional information about a field (such as warnings and error messages) is now a label for the input of that field.

In addition to these changes, a couple of specific changes were made to individual setting fields:
- The bubble position setting is now a simple select input.
- The template position label was broken into horizontal and vertical parts.

Finally, `button-bar`s were changed from being generic to using lists. This change was incited by requiring an element allowed inside of paragraph elements be used. `ul` worked for this and also holds some useful semantic meaning given that `button-bar`s are flat, unordered sets of usually two or more buttons.

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -4921,7 +4921,7 @@ window.App = (function() {
             lblUsernameColor,
             lblIgnores,
             lblIgnoresFeedback
-          ].map(x => crel('p', x))
+          ].map(x => crel('div', x))
         );
         modal.show(modal.buildDom(
           crel('h2', { class: 'modal-title' }, 'Chat Settings'),
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -4766,21 +4766,10 @@ window.App = (function() {
           return [cb, lbl];
         };
 
-        const a24hTimestamps = genCheckboxGroup('24 Hour Timestamps');
-        const _cb24hTimestamps = a24hTimestamps[0];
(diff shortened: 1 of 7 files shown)


resources/public/pxls.js
```diff
@@ -4755,57 +4755,79 @@ window.App = (function() {
...
+            ' ',
+            crel('span', { class: 'label-text' }, label));
+          return [cb, lbl];
+        };
 
-        const _cb24hTimestamps = crel('input', { type: 'checkbox' });
-        const lbl24hTimestamps = crel('label', _cb24hTimestamps, '24 Hour Timestamps');
+        const a24hTimestamps = genCheckboxGroup('24 Hour Timestamps');
+        const _cb24hTimestamps = a24hTimestamps[0];
+        const lbl24hTimestamps = a24hTimestamps[1];
```
netux, line 4771: Could use
```js
const [ _cb24hTimestamps, lbl24hTimestamps ] = genCheckboxGroup('24 Hour Timestamps');
```
here and avoid some repetition.

BlankParenthesis: I initially avoided using destructuring on purpose since it isn't really used elsewhere but if we are to move to more modern JS standards, you are correct that this would look a lot nicer.

PR #369, opened 2020-06-02
personal info rule and profile link

cp-codes2:
In separate pr per flying's request. Deleted some extra white space as well. Noticed that profile link wasn't in there so I added it too.


FlyingSixtySix:
I felt that the rule would be better kept short and to the point, so I just shortened it to "no personal information." It can be handled case-by-case.

PR #371, opened 2020-06-03
Donator role and badge

cp-codes2:
will have to test everything. at the moment I have no way of doing so until I figure out how to run an instance of pxls to test the changes. only went through a bit so chances are things need to be added to make it work properly. The idea came from a pxls clone (stem place) but the code for the role was lost a few months ago so I am remaking it but for pxls. I'm happy to have help if needed.

### DIFF
--- a/src/main/java/space/pxls/user/User.java
+++ b/src/main/java/space/pxls/user/User.java
@@ -241,6 +241,14 @@ public class User {
                 roleName = Character.toUpperCase(roleName.charAt(0)) + roleName.substring(1);
                 toReturn.add(new Badge(roleName, roleName, "icon", "fas fa-shield-alt"));
             }
+        if (getRole() != null && getRole().greaterEqual(Role.TRIALMOD)) {
+            if (getRole().equals(Role.DONATOR)) {
+                toReturn.add(new Badge("Donator", "Donator", "icon", "fas fa-dollar-sign"));
+            } else if (getRole().greaterEqual(Role.TRIALMOD)) {
+                String roleName = getRole().name().toLowerCase();
+                roleName = Character.toUpperCase(roleName.charAt(0)) + roleName.substring(1);
+                toReturn.add(new Badge(roleName, roleName, "icon", "fas fa-shield-alt"));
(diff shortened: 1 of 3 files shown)


src/main/java/space/pxls/user/User.java
```diff
@@ -236,6 +236,8 @@ public String getTag() {
         if (getRole() != null && getRole().greaterEqual(Role.TRIALMOD)) {
             if (getRole().equals(Role.DEVELOPER)) {
                 toReturn.add(new Badge("Developer", "Developer", "icon", "fas fa-wrench"));
+            if (getRole().equals(Role.DONATOR)) 
+                toReturn.add(new Badge("Donator", "Donator", "icon", "fas fa-dollar-sign"));
```
cp-codes2, line 240: fixed obvious issues. now just have to test and see if I there was something I overlooked.

netux:
I think DONATOR shouldn't be it's own role, but rather a column on the database.
This way staff members and banned users can be tagged as donators, and given a separate badge in chat.

cp-codes2:
hmm I never thought about how I would work with staff and more specifically users with the banned role. Quite honestly im not the person to make that change. If the pr ends up not being merged due to that fact, maybe someone would want to re approach donator that way?

BlankParenthesis:
While adding a single flag for donator rather than role does seem a lot simpler and flexible than adding a role, I don't think it addresses the real problem.

Any role currently implemented could be done with the same method - there's very little that makes a role mutually exclusive to all other roles. If the project plans on having more roles and role-like additions in future, it might be worth admitting the hard truth that a significant rework to roles might be desirable such that a user can hold multiple at a time.

This would allow developers to also be moderators without or, more importantly, allow developers to not also be granted moderator privileges. It would also allow for more decorative roles similar to Discord, if desired.

FlyingSixtySix:
What @BlankParenthesis pointed out is correct; the underlying system of roles needs to be addressed before moving forward. Thank you for your work on the pull request--it at least brought us closer to a better system!

PR #378, opened 2020-06-07
separate clearing for heat map and virginmap

cp-codes2:
noticed a suggestion in dev and bugs for this. sounds like a good idea to make them clear separately.

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -2084,7 +2084,7 @@ window.App = (function() {
               return;
             }
 
-            if (evt.key === 'o' || evt.key === 'O' || evt.which === 79) {
+            if (evt.key === 'o' || evt.key === 'O' || evt.which === 79 ||evt.which === 111) {
               heatmap.clear();
             }
           });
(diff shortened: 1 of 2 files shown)


resources/public/pxls.js
```diff
@@ -2132,7 +2132,7 @@ window.App = (function() {
               return;
             }
 
-            if (evt.key === 'o' || evt.key === 'O' || evt.which === 79) { // O key
+            if (evt.key === 'u' || evt.key === 'U' || evt.which === 79) { // U key
```
netux, line 2135: Key code for 'U' is 85, for 'u' it is 117.

netux: While you're at it, can you add a check for when `evt.which === 111` on the one that checks for O (heatmap clear)?

netux:
You missed updating the Keybinds tab under Settings
image

cp-codes2:
👍 always miss stuff like that :notlikethis: will do

cp-codes2:
ok everything thus far has been fixed though `evt.which === 111` has not yet been added

PR #379, opened 2020-06-09
Remove backup rules for CSS Variables

BlankParenthesis:
When CSS Variables were implemented, the default style colors were left declared before the new variables were. This gave fallback support if a user's browser did not support CSS variables.

This removes that fallback.

I haven't noted any complaints about themes not working, which would be a symptom of users not having support for CSS variables, making me confident in making this PR. 

This PR can be delayed until it is deemed sufficiently safe.

### DIFF
--- a/resources/public/style.css
+++ b/resources/public/style.css
@@ -156,9 +156,7 @@ html, body {
     margin: 0;
     width: 100%;
     height: 100%;
-    color: #000;
     color: var(--general-text-color);
-    background: #CDCDCD;
     background: var(--general-background);
     overflow: hidden;
     position: relative;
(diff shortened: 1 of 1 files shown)


netux:
You left one on line 1463
image

netux:
As for the mergeability of the PR, while I haven't heard any complaints either but I'd still wait until [Can I Use says CSS Variables are supported by 95%+ of users](https://caniuse.com/#feat=css-variables), just to be sure.

BlankParenthesis:
It seems like emoji-button is using cssvars now (not sure if there are backups there though). Perhaps that can be persuasive.

FlyingSixtySix:
I think we're close enough on support for it to affect only a tiny minority of users.

PR #389, opened 2020-06-22
Role Overhaul

FlyingSixtySix:
**Database Migration**

```sql
-- Create the new roles table
CREATE TABLE roles (
    id INTEGER REFERENCES users,
    role VARCHAR(512),
    PRIMARY KEY (id, role)
);

-- Add the new shadow-banned column
ALTER TABLE users ADD COLUMN is_shadow_banned BOOL NOT NULL DEFAULT false;

-- Migrate specific roles

-- USER
-- Assuming we're using the default roles in the PR, no need to assign a default role in the database.
--INSERT INTO roles SELECT id, 'user' FROM users WHERE role = 'USER';

-- BANNED
UPDATE users SET ban_expiry = to_timestamp(0) WHERE role = 'BANNED';

-- SHADOWBANNED
UPDATE users SET is_shadow_banned = true WHERE role = 'SHADOWBANNED';

-- TRIALMOD
INSERT INTO roles 'trialmod' FROM users WHERE role = 'TRIALMOD';

-- MODERATOR
INSERT INTO roles 'moderator' FROM users WHERE role = 'MODERATOR';

-- ADMIN
INSERT INTO roles 'administrator' FROM users WHERE role = 'ADMIN';

-- DEVELOPER
INSERT INTO roles 'developer' FROM users WHERE role = 'DEVELOPER';

-- Remove now-unused role column
ALTER TABLE users DELETE COLUMN role;
```
<hr>
This PR overhauls our role system in its entirety. There are way too many individual changes to note, but here's a summary to the best of my abilities:

- Instead of being limited to a single role, users can now have multiple roles.
- Roles are now objects with an ID, name, type, default flag, inheritance, badges, and permissions.
- Roles are now configurable via `roles.conf`. When setting up an instance, users will now need to copy `resources/roles-reference.conf` to `roles.conf`.
- Ban logic has been reworked to where `ban_expiry` being `0` (`1970` timestamp) counts as a permanent ban, `null` counts as the user not being banned, and a future timestamp counting as a regular timed ban. Because of this, and a lack of need, there is no longer a `banned` role.
- Shadow-bans have been reworked as a `is_shadow_banned` column rather than a role.
- Role badges have been delegated to the configuration file.
- Permissions are string-based "nodes," such as `board.data`. `roles.md` details all known permission nodes, the endpoints they're found on, their purposes, and the default role (based on `roles-reference.conf`).
- Console commands relating to role and ban management have been polished.


src/main/java/space/pxls/user/Permission.java
```diff
@@ -0,0 +1,13 @@
+package space.pxls.user;
+
+public class Permission {
```
BlankParenthesis, line 3: This entire class is just an abstraction of a `String`. It can be safely factored out I suspect. A more concrete `Permission` type would probably be an `enum`.

src/main/java/space/pxls/user/PermissionManager.java
```diff
@@ -0,0 +1,40 @@
+package space.pxls.user;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class PermissionManager {
```
BlankParenthesis, line 8: Since `Permission` is just a `String`, this class becomes a singleton for hash-map of strings that map to themselves. Furthermore, any retrieval returns the input and stores if it it does not already exist on the map. None of this does anything. This class can be removed.

src/main/java/space/pxls/user/RoleManager.java
```diff
@@ -0,0 +1,48 @@
+package space.pxls.user;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class RoleManager {
```
BlankParenthesis, line 6: Perhaps this class would be better as a static constructor on `Role`. Something like `.fromID`. `Role` could store the same hash-map that this does but as a static field which is more representative of the singleton-like usage of `Role`s.

src/main/java/space/pxls/user/User.java
```diff
@@ -98,8 +101,10 @@ public int getId() {
 
     public boolean canPlace() {
         if (isRenameRequested) return false;
-        if (role.greaterEqual(Role.MODERATOR) && overrideCooldown) return true;
-        if (!role.greaterEqual(Role.USER) && role != Role.SHADOWBANNED) return false; // shadowbanned seems to be able to place
+
+        if (hasPermission("board.cooldown.override") && overrideCooldown) return true;
+        if (!hasPermission("board.place")) return false;
+        if (hasPermission("board.place.localonly")) return false; // Shadowban logic
```
BlankParenthesis, line 107: This check taking place after just checking that we have `board.place` indicates that it is possible to have `board.place` and `board.place.localonly`. If we take a strict view of permissions, they should *only* grant new permissions. Here, it is clear that `board.place.localonly` is only used to deny certain features of `board.place` rather than granting only enough power to place pixels locally. Clearing this up will probably make code elsewhere more readable too.

src/main/java/space/pxls/server/WebHandler.java
```diff
@@ -798,7 +798,7 @@ public void chatban(HttpServerExchange exchange) {
             sendBadRequest(exchange);
             return;
         }
-        if (user.isBanned() || user.getRole().lessEqual(Role.USER)) {
+        if (user.isBanned() || user.isBanned()) {
```
BlankParenthesis, line 801: Redundant check to `isBanned()`

src/main/java/space/pxls/user/User.java
```diff
@@ -158,12 +164,31 @@ public boolean updateCaptchaFlagPrePlace() {
     }
 
     public void setOverrideCooldown(boolean overrideCooldown) {
+        if (!hasPermission("board.cooldown.override")) overrideCooldown = false;
```
BlankParenthesis, line 167: This function can just set `this.overrideCooldown` to `hasPermission("board.cooldown.override") && overrideCooldown` which is much clearer.

src/main/java/space/pxls/user/User.java
```diff
@@ -367,64 +426,39 @@ public void chatban(Chatban chatban) {
         return connections;
     }
 
-    public void shadowban(String reason, int rollbackTime, User banner) {
+    public void shadowBan(String reason, int rollbackTime, User banner) {
         setBanReason(reason);
-        setRole(Role.SHADOWBANNED, true);
```
BlankParenthesis, line 372: Thanks to the removal of this, shadowban status is no longer saved to the database as far as I can tell. This means that a shadowban only lasts until the server restarts.

src/main/java/space/pxls/user/Permission.java
FlyingSixtySix: I implemented it just in case we wanted to add anything extra to how permissions are represented. I'll take it out.

src/main/java/space/pxls/user/PermissionManager.java
FlyingSixtySix: This was also another just-in-case type of thing. I'll take it out.

src/main/java/space/pxls/user/User.java
FlyingSixtySix: The `board.place.localonly` permission node is a left-over from before I decided to make shadow-bans based on a column. I'll clean this up.

src/main/java/space/pxls/server/WebHandler.java
FlyingSixtySix: Don't know how I missed that!

src/main/java/space/pxls/user/User.java
FlyingSixtySix: Looks like I forgot to implement shadow-banning on the database! I'll take care of that.

src/main/java/space/pxls/App.java
```diff
@@ -26,13 +26,17 @@
...
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class App {
     private static Gson gson;
     private static Config config;
     private static Database database;
     private static UserManager userManager;
+    private static List<String> permissions;
```
BlankParenthesis, line 39: As far as I can tell, this is never used.

cp-codes2:
it doesn't appear read me has been changed in this pr. I suggest a change describing roles.conf in read me.

src/main/java/space/pxls/user/User.java
```diff
@@ -281,26 +329,42 @@ public String getTag() {
         return toReturn;
     }
 
-    public boolean isPermaBanned() {
-        return this.role == Role.BANNED;
+    public boolean isBanned() {
+        if (banExpiryTime != null && banExpiryTime > System.currentTimeMillis()) {
+            // If now is past their ban expiry, erase their ban expiry time.
+            banExpiryTime = null;
+            App.getDatabase().updateBan(this, null);
```
netux, line 336: I don't know if I like the idea of having a function that is expected to simply check something update the user as well. Maybe instead of checking this and updating the db we just `return banExpiryTime != null && banExpiryTime > System.currentTimeMillis();`

netux: The update to the database can be done elsewhere (or maybe never, as the suggestion above should solve the problem of code only checking `banExpiryTime != null`)

src/main/java/space/pxls/data/DBPixelPlacement.java
```diff
@@ -75,12 +74,12 @@ public DBPixelPlacement map(ResultSet r, StatementContext ctx) throws SQLExcepti
                     r.getInt("u_id"),
                     r.getString("username"),
                     r.getString("login"),
-                    Role.valueOf(r.getString("role")), // TODO: all users might not have valid roles
+                    App.getDatabase().getUserRoles(r.getInt("u_id")),
```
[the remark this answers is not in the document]
FlyingSixtySix: I would rather serve role objects on an endpoint (ideally within `/info`) and serve role IDs if wanted in the future. For now, I'll comment this out.

PR #391, opened 2020-07-08
Implement Donator Role/Badge/Gradient

cp-codes2:
this is my attempt at adding donator role with a unique green donator gradient as a small thank you. The gradient is a cycle through the 3 green colors on the palette and is under the id -2. Here is an css example of how it should look in game including the badge
Screen Shot 2020-07-07 at 7 51 27 PM
I feel like I'm missing something with the perm "chat.usercolor.donator" but haven't found anything yet so this will need some testing.
this will also require a few changes for the subdomains of Pxls so expect a pr for those changes soon unless something changes


cp-codes2:
so testing this on stem place reveals how insanely broken this is so gonna go a complete overhaul of this code then I will remove it from a draft

FlyingSixtySix (approved):
I made a small mistake by adding some recently merged commits to the PR, but all-in-all, looks good to me! I also fixed a couple of things like missing `isDonator()` in the return for `user`, and lowing the namecolor boundary to -2. We should find a better solution later, as future namecolors that go negative would need to drop this boundary even more.

PR #400, opened 2020-07-30
New "matte" theme

FeeeeK:


resources/public/pxls.js
```diff
@@ -4817,19 +4830,18 @@ window.App = (function() {
         // dom generation
         const body = crel('div', { class: 'chat-settings-wrapper no-p-margin' });
 
-        const genCheckboxGroup = (label) => {
-          const cb = crel('input', { type: 'checkbox' });
-          const lbl = crel('label', { class: 'input-group' },
-            cb,
+        const genCheckboxGroup = (label, id) => {
```
BlankParenthesis, line 4833: Many of the changes in the PR stem from this change. It changes the current hierarchy of nodes from having the input contained within the label, to having the input precede the label and reference it with the `for` attribute. Both of these are valid ways to associate a label with an input, as mentioned in the [MDN article on the label element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label).

When doing a recent restructure of the HTML in the settings panel, I chose to keep the current method of having the input be a child of the label. My reasoning was that it was more flexible as unique ids need not be generated for every input and also happened to be the existing system, with no need for change besides preference.

That said, I would request that the development team come to a conclusion on which form the project wants to use from now on and stick to it before merging this PR.

resources/public/pxls.js
```diff
@@ -3397,6 +3401,15 @@ window.App = (function() {
...
         });
 
+        $('input[type=range]').on('input', function(e) {
+          var min = e.target.min;
+          var max = e.target.max;
+          var val = e.target.value;
+          $(e.target).css({
+            backgroundSize: (val - min) * 100 / (max - min) + '% 100%'
+          });
+        }).trigger('input');
```
netux, line 3411: Since this triggers _all_ input handlers it causes and obscure bug where the view won't be set on the center of the canvas when the page is loaded without any query arguments on the URL.

From my debugging:
- The page is centered inside [`board.start()`](https://github.com/pxlsspace/Pxls/blob/ddb590a9c9a5dc9d1de24f4be7c500f0bb2fdb23/resources/public/pxls.js#L1549), where it gets the coordinates by doing `query.get('x') || self.width / 2` and `query.get('y') || self.height / 2`. This depends on query parameters not being initialized yet (thus `query.get('x')` and `query.get('x')` would return undefined)
- This trigger to all input event handlers
  1. triggers the event handler for [template opacity slider, which calls `template.update()`](https://github.com/pxlsspace/Pxls/blob/ddb590a9c9a5dc9d1de24f4be7c500f0bb2fdb23/resources/public/pxls.js#L2429)
  2. [which calls `board.update()`, where the query is updated with values that default to 0](https://github.com/pxlsspace/Pxls/blob/ddb590a9c9a5dc9d1de24f4be7c500f0bb2fdb23/resources/public/pxls.js#L1617)
  3. thus setting the query x and y to 0 before they are presumed to be undefined.

There are multiple points in which this can be fixed, but maybe the one that touches less parts of the old code is to only run the new event handler that updates the color of the input bar manually instead of triggering all input handlers.

resources/public/pxls.js
netux: Either style is okay for me, but I'd rather keep it consistent and nest the input inside the label like how it was before.
Is there a way to re-implement the new visuals that way?

resources/public/pxls.js
```diff
@@ -4817,19 +4830,18 @@ window.App = (function() {
         // dom generation
         const body = crel('div', { class: 'chat-settings-wrapper no-p-margin' });
 
-        const genCheckboxGroup = (label) => {
-          const cb = crel('input', { type: 'checkbox' });
-          const lbl = crel('label', { class: 'input-group' },
-            cb,
+        const genCheckboxGroup = (label, id) => {
+          const cb = crel('input', { id: id, type: 'checkbox' });
+          const lbl = crel('label', { for: id, class: 'input-group' },
```
netux, line 4835: Adding to the thing above, this element is even inconsistent with the one on the settings panel, and causes a bug where the new checkbox doesn't display for chat settings.
image

resources/public/themes/matte.css
```diff
@@ -0,0 +1,298 @@
...
+
+input[type="checkbox"]+span {
+    display: block;
+    position: relative;
+    padding-left: 20px;
+    cursor: pointer;
+}
+
+select {
+    -webkit-appearance: none;
```
netux, line 192: On Firefox this has the effect of hiding the dropdown arrow to the right of `<select>`s.
image
This made it harder for me to identify what the thing was. Please consider removing this rule.
image

netux (changes requested):
Team looks stylish but needs some changes and bug fixes.

resources/public/pxls.js
```diff
@@ -3442,6 +3450,27 @@ window.App = (function() {
...
+          var val = e.value;
+          $(e).css({
+            backgroundSize: (val - min) * 100 / (max - min) + '% 100%'
+          });
+        }
+        ranges.on('input', (e) => updateBar(e.target));
+        ranges.each((idx, element) => updateBar(element));
+      },
+      prettifySelect: function (select) {
+        $(select).click(function () { $(this).toggleClass('select-arrow'); });
```
BlankParenthesis, line 3467: This is insufficient to keep track of the `select`'s  state.
As an example, a user could click on the select to open the dropdown, then click elsewhere on the page to close it.
This leaves the arrow in the wrong orientation.

One alternative is instead binding the arrow the the `:focus` selector in CSS, though that too has some state issues - albeit probably fewer and ones less persistent.

resources/public/themes/matte.css
```diff
@@ -1,3 +1,5 @@
+# "Matte" theme by github.com/feeeek
```
netux, line 1: This is not how you add comments in CSS. I'm guessing you meant
```css
/* # "Matte" theme by github.com/feeeek */
```

This is the cause of your theme's colors breaking.
image

With that fixed:
image

resources/public/pxls.js
netux: I agree with Blank, but I'd suggest just leaving the arrow pointing down all the time for now.

netux (changes requested):

netux:
Your "Fix merge conflicts" commit didn't merge all the changes from master to this branch.
For example, the commit updates index.html to load pxlsMarkdown.js instead of anchorme, but pxls.js still references anchorme (which breaks chat).

Please correctly merge master into this branch.

netux (approved):

resources/public/pxls.js
```diff
@@ -3404,6 +3410,8 @@ window.App = (function() {
         self._initAccount();
         self._initBanner();
         self._initMultiTabDetection();
+        self.prettifyRange('input[type=range]');
+        self.prettifySelect('select');
```
netux, line 3414: You forgot to remove this.

netux (changes requested):

netux (approved):

PR #402, opened 2020-08-06
changes to the info/rules interface

cp-codes2:
added some spacing between lines so it does not look like a giant wall of text as much

### DIFF
--- a/resources/public/info.html
+++ b/resources/public/info.html
@@ -46,7 +46,7 @@
 	</header>
 	<div class="pad-wrapper">
 		<ol>
-			<li><p>Keep chat civil. No harassment or homophobic/transphobic language.</p></li>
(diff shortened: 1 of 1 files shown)


BlankParenthesis:
The adding of paragraphs to each item is semantically questionable, complicates the DOM, and - in the case of contained sub-lists - is incorrect. 

This would probably be better implemented in CSS which is the appropriate place to implement purely visual changes such as this. This could be as a general change to the display of ordered list items or using a class.

FlyingSixtySix (changes requested):
I agree with @BlankParenthesis; with the goal of this being added spacing (as far as I understand), it would be more efficient to handle it via CSS than more elements.

cp-codes2:
Will do. Haven't had much time to work on this lately but shouldn’t take long

FlyingSixtySix:
Closed due to inactivity.

PR #404, opened 2020-08-19
Palette configuration overhaul

netux:
- Move palette configuration into palette.conf
- Allow colors to have a name
  - Default names taken from PxlsFiddle.com
- The `palette` field returned by the /info endpoint is now a list of objects with fields `name` and `value` ("RRGGBB" string)
- pxls.js was reorganized to utilize the new color names, along with some other miscellaneous changes
- The palette is now internally a class that holds a list of Color and the index of the default/background color

#### Breaking changes
This PR moves palette information into its own file, palette.conf. The palette must be migrated to this new file for Pxls to work.


resources/palette-reference.conf
```diff
@@ -0,0 +1,122 @@
+colors: [
+	// Each color must follow the following format:
+	// {
+	// 	  name: "my color" // public name for the color
+	//    value: HEX // color value, in RGB HEX
+	//	  background: false // optional, whenever the color is used as background in the default board
```
BlankParenthesis, line 6: Would it be possible to instead have this be a separate root-level key?
As it is, this seems to have an external condition that it appear on exactly one item in this array - something which can be avoided by moving it to a separate key and internalizing the condition with the structure.
This key could specify an index or name, whichever seems more appropriate (or perhaps allow either).

resources/public/pxls.js
```diff
@@ -1543,13 +1548,12 @@ window.App = (function() {
           self.centerOn(cx, cy, true);
           socket.init();
 
-          (async function() {
```
BlankParenthesis, line 1546: I believe that by removing this, the app will hang until this data is fetched which sounds undesirable..

netux: The data is still fetched asynchronously because we still `await binaryAjax` . The site will hang either way when processing and drawing the board, but that's a problem from before.
I just removed this function and made the upper function (callback to fetching /info) async instead.

resources/palette-reference.conf
netux: I tried to avoid having to worry about indices, especially with the list now spanning multiple lines, which is why I chose to use a `background` key inside the color (similar to how in roles.conf the default role gets a `default: true` attribute).

Your idea of supporting names as well as the index works too, but it introduces an edge case of a color being set as background, but then deleted. This way we make sure an unknown color cannot be set as background.

I handled both whenever a background is never set (defaults to the first color in the list) or a background is set more than once (throws an exception on boot).

resources/public/pxls.js
BlankParenthesis: Having tested this, the only noticeable difference is that the template will no longer load while waiting for the board data. This is because the hash params don't get queued until later in this function. This can be resolved by moving this section to the end of this function; something which seems reasonable given it's the only async code.

Titeiiko:
this man spitting facts here

resources/palette-reference.conf
Titeiiko: this man do be spitting facts

PR #407, opened 2020-08-19
Ask for confirmation when moving a template with the "Move Template Here" button

netux:


BlankParenthesis:
This implementation looks fine, but I don't like the general practice of asking for confirmation of a clear action. Additionally, this is not a destructive action and can easily be undone by navigating backwards in browser history.

netux:
This was targeted at mobile users who might fat-finger the button by accident, as I've been told moving the template back is a pain.
Maybe adding a checkbox to never ask again (and a mirror checkbox on Settings) would make it less intrusive, at least for the subsequent times?

BlankParenthesis:
That strikes me as a bad solution to another bad solution - if pressed, I would prefer we only have one bad solution rather than a series of stacked ones.

If the button is used so infrequently that anyone clicking the button is likely to have done so by mistake, perhaps we should look instead at removing it or replacing it with a better solution.

netux:
Can't disagree there. Closing this PR to get to work on another system for it.

PR #410, opened 2020-08-25
Virginmap and Heatmap overlay excluding non-placeable pixels

Mikarific:
Sorry about the excess commits... I was trying to figure out why my commits weren't linking with my github account. 

Anyways, this commit prevents any changed pixels outside of the placemap from rendering at all on the heatmap and virginmap, and renders the heatmap and virginmap backgrounds only on the placeable pixels. This is done by adding two canvases for each background, as it was the most efficient solution without any drawbacks (such as removing the background opacity setting in settings).

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -1765,7 +1765,9 @@ window.App = (function() {
 
         self.elements.board.toggleClass('pixelate', scale > 1);
         overlays.heatmap.setPixelated(scale >= 1);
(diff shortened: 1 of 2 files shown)


BlankParenthesis:
This looks good. My one complaint would be that having the background and foreground be treated as different overlays seems like clutter at best and incorrect at worst - I would suggest you modify `overlay` to manage both the foreground and background rather than creating an `overlay` for each.

PR #417, opened 2020-08-28
Virginmap and Heatmap overlay excluding non-placeable pixels

Mikarific:
Old PR came from my master branch because I didn't think I'd be making more PRs and I was lazy...

Anyways, this commit prevents any changed pixels outside of the placemap from rendering at all on the heatmap and virginmap, and renders the heatmap and virginmap backgrounds only on the placeable pixels. This is done by adding two canvases for each background, as it was the most efficient solution without any drawbacks (such as removing the background opacity setting in settings).

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -1765,7 +1765,9 @@ window.App = (function() {
 
         self.elements.board.toggleClass('pixelate', scale > 1);
         overlays.heatmap.setPixelated(scale >= 1);
+        overlays.heatbackground.setPixelated(scale >= 1);
         overlays.virginmap.setPixelated(scale >= 1);
+        overlays.virginbackground.setPixelated(scale >= 1);
         template.setPixelated(scale >= template.getWidthRatio());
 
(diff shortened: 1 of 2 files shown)


BlankParenthesis:
I'll just quote my comment on the previous issue then:
> This looks good. My one complaint would be that having the background and foreground be treated as different overlays seems like clutter at best and incorrect at worst - I would suggest you modify `overlay` to manage both the foreground and background rather than creating an `overlay` for each.

netux (commented):
I agree with Blank's suggestion. Using separate overlays is okay for a first implementation, but generalizing the `overlay` module to support multiple layers would be ideal in the future.

netux:
I also noticed a bug where if a pixel is updated outside of bounds (due to a nuke, for example), the heatmap/virginmap for those pixels is still drawn. Adding a check for placeability of the section when updating the overlays would solve the issue, though it's low priority since, in practice, nukes outside the placemap are rare.

PR #421, opened 2020-09-09
Fix broken color brightness workaround

BlankParenthesis:
This should make the workaround more reliable.

Included is a change of the element used to maybe or maybe not improve performance 🤷.


netux (approved):
I noticed that toggling an overlay on and off briefly makes the canvas blurry, but this is a good improvement. Good job.

PR #434, opened 2020-11-05
Store chat message purge reason and allow staff to retrieve purged messages on chat history.

netux:
This PR allows users with the `chat.history.purged` permission to retrieve purged chat messages from the ChatHistory packet exchange.
The new `purge_reason` column on the database was necessary to return the ban reason from chat history.

This PR also makes the server action message sent when a specific purge (message delete) is done appear for everyone, and not just the user who's message was purged.

### Breaking change
This PR adds a new field to the `chat_messages` table on the database. The following query must be run on the database before deployment:
```sql
ALTER TABLE chat_messages ADD COLUMN purge_reason TEXT;
```
Sadly it is not possible to populate the column with old values.


FlyingSixtySix (changes requested):
Initial chat messages show up as white. LGTM otherwise!

PR #435, opened 2020-11-07
Expose default board as /initialboarddata

netux:
Exposes the default_board.dat file as an endpoint on the server. This allows third party developers to retrieve this information at any time on the canvas and use it for their own applications.

For example, https://pxlsfiddle.com can benefit from this by automating the default_board.dat retrieval process, since as far as I know the initial state is uploaded manually every new canvas from the image posted on our Discord.

### DIFF
--- a/src/main/java/space/pxls/App.java
+++ b/src/main/java/space/pxls/App.java
@@ -751,6 +751,10 @@ public class App {
         return board;
     }
 
+    public static byte[] getDefaultBoardData() {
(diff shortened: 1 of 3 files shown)


Undoundoundo:
While pxlsFiddle does automatically grab the new board, due to the fact that it only polls every few minutes it does mean that it can be a little late to the party and pixels will already be placed - so yes, this would be nice to have :)

PR #438, opened 2020-11-18
Added new theme "Terminal"

SoftwareKitten:
Darker + green theme built off of matte

### DIFF
--- a/resources/public/pxls.js
+++ b/resources/public/pxls.js
@@ -3408,6 +3408,11 @@ window.App = (function() {
           name: 'Matte',
           location: '/themes/matte.css',
           color: '#468079'
+        },
+        {
+          name: 'Terminal',
+          location: '/themes/terminal.css',
+          color: '#94e044'
         }
       ],
       specialChatColorClasses: ['rainbow', 'donator'],
--- a/resources/public/themes/terminal.css
+++ b/resources/public/themes/terminal.css
@@ -218,7 +218,7 @@ input[type="checkbox"]+span:before {
     display: block;
     width: 12px;
     height: 12px;
-    border: 1px solid #004e00;
+    border: 1px solid #007300;
     position: absolute;
     left: 0;
     top: 2px;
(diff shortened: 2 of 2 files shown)


cp-codes2:
looks really good to me
couple things I noticed though are the dark green text blending in with black at some places like 
Screen Shot 2020-11-18 at 10 57 36 AM
Screen Shot 2020-11-18 at 11 00 45 AM

and maybe the border around palette colors 
Screen Shot 2020-11-18 at 10 59 17 AM

other than I that I think its pretty much ready to go
nice work

SoftwareKitten:
I made those dark green on purpose
having it all light green kinda looked bad, I wanted to break it up a bit

netux:
I agree with CP. Some dark green text is unreadable for some (including me).

Since you don't like light green, may I suggest using white maybe?

SoftwareKitten:
hmm
mabye i can make it slightly lighter

SoftwareKitten:
there we go... it is lighter, but not as light as the rest of the light green
i also made the panel close button lighter... because I realized that was bugging me that i couldnt see it well... hope that is good!

SoftwareKitten:
hmmm only thing i would want to change... which i don't think I CAN change... is to make the flashing "|" typing indicator a flashing underscore
but I don't think I can change that with css

netux:
The color for toggled checkmarks is very dark in contrast with the background, almost invisible.
image

You still preserve the slider tick color from the Matte theme. Is this intentional?
image

The palette borders still look weird to me
image


Overall it's a nice theme, though the contrast is awful.
I get that that's the point, but I wouldn't use it too much 😅.

SoftwareKitten:
its terminal
huh... i thought i fixxed the slider... hmmmmmmm.... yeah i fix checkboxes in a sec
i tried to make black background a very dark green..... so idk what you meany by contrast is aweful... its just a dark mode with cool green accents

SoftwareKitten:
i copy pasted the text file into my inspect element and slider is green
mabye i messed up in github hold on

SoftwareKitten:
no it shouldn't still be matte theme

SoftwareKitten:
changed palette border colors
this is the final change (unless find something else after its released)

PR #440, opened 2020-11-25
Reorder settings so that they are easier to navigate.

netux:
The "General Settings" section was split into UI Settings, Overlay Setting, Control Settings, Snapshot Settings and Sound and Notification Settings (extended from Pixel Alert Settings).
The settings were also sorted so that the most useful ones (in my opinion) stay at the top of their respective sections.

image
image
image


BlankParenthesis:
Nice to see someone finally got around to this.
Personally I would also consider moving chat settings into the settings panel - the separation is a bit strange and a new "chat" section would be the perfect place to add a "enable chat" toggle that some seem to want.

The chat settings themselves probably need some help but I'm not sure if this is right PR for tidying that up.

netux:
I have thought about that, but I don't like the idea of making the user close the chat panel to open chat settings. Sounds bit counterintuitive imo.

BlankParenthesis:
Having to close the settings panel to visit chat settings seems worse to me. Having two separate systems also increases maintenance, makes UX familiarity harder and means we have to make a tough choice about where each setting goes every time one is related to chat. For example - as mentioned in my previous comment - where would you put an "enable chat" toggle? The reality is that you can't put it in the chat panel if it disables that panel so it *must* go in the regular settings panel - separate from all the other chat-related settings.

netux:
Been prototyping it for these past few days and got it to something pretty-ish 
image
I also rerouted the cog in chat to open the settings panel and scroll to the Chat Settings section.

What do you think?

BlankParenthesis:
Looks good to me. Only concern is that the ping options also fit into the sound category.

resources/public/pxls.js
```diff
@@ -664,6 +664,16 @@ window.App = (function() {
     });
 
     return {
+      // utilities
+      filter: {
```
netux, line 668: I realize this adds an unrelated field to the `settings` export, but it was the quickest solution I could think of without having to move the filter code somewhere else (like `uiHelper`).
Moving this can be done on another PR.

resources/public/pxls.js
```diff
@@ -6685,7 +6592,6 @@ window.App = (function() {
           } else {
             self.elements.userInfo.fadeIn(200);
           }
-          self.roles = data.roles;
```
netux, line 6688: Moved this up so that populating the username color select on `pxls:user:loginState` is role aware (rainbow and donator options appear correctly).

PR #450, opened 2020-12-17
Add a small note explaining that the keybinds are based on the QWERTY layout

netux:
image


FlyingSixtySix:
I think this is fine for now; in the future we might want to add alt text or something similar by the header to take up less vertical space.

PR #451, opened 2020-12-19
Fix `Unexpected token '.'` error on Safari.

Mikarific:


FlyingSixtySix:
It's a shame my fix didn't actually fix the issue, even though the test case I set up passed... something to figure out another time.

PR #452, opened 2020-12-19
Add configurable custom emoji

Mikarific:
This PR relies on the changes made on [pxls-markdown](https://github.com/pxlsspace/pxls-markdown/pull/4) and [emoji-button](https://github.com/netux/emoji-button/pull/1). The compiled result of both of those PRs are included in this one.

Changes that must be made for this PR to work:
- Create `emoji` directory in the storage directory.
- Create a `customEmoji` list in `pxls.conf` like the one below.
```
customEmoji: [
  { name: "kek", emoji: "kek.png" },
  { name: "TrueTrue", emoji: "truetrue.png" }
]
```
- Create files with the same name as the above configured names in your new created `emoji` directory.

Things to note:
- Emoji names have only been tested to support `a-z`, `A-Z`, `_`, and `-`.
- No two emoji may have the same name, this is case insensitive, so `TrueTrue` cannot coexist with `truetrue`, but they can exist seperately.
- This PR reorganizes when typeahead, the emoji picker, and a few other things are initialized in order to get the custom emoji info from the `/info` endpoint before typeahead and the emoji picker are created.


src/main/java/space/pxls/server/WebHandler.java
```diff
@@ -1714,7 +1716,8 @@ public void info(HttpServerExchange exchange) {
             Math.min(App.getConfig().getInt("chat.characterLimit"), 2048),
             App.getConfig().getBoolean("chat.canvasBanRespected"),
             App.getConfig().getStringList("chat.bannerText"),
-            App.getConfig().getBoolean("oauth.snipMode")
+            App.getConfig().getBoolean("oauth.snipMode"),
+            emoji
```
netux, line 1720: I think it's better to read the `chat.customEmoji` list on each request like the rest of these parameters are doing. That way the emotes are updated when the config is reinitialized (e.g. from a `reload`) and we don't have to restart the server to update them.
```suggestion
            App.getConfig().getList("chat.customEmoji").unwrapped();
```

netux (changes requested):
Small nitpicks here and there, good job overall.

FlyingSixtySix (approved):
Works as expected and did not find any bugs. LGTM!

PR #456, opened 2020-12-24
Implement template styling in WebGL

BlankParenthesis:
This adds template styling similar to that on [pxlsfiddle](https://pxlsfiddle.com) which can be efficiently rendered in real-time thanks to WebGL shader programs.

This makes it much easier for a user to work on a template using their own personally preferred style rather than relying on either re-processing the template themselves or trusting the template owner to provide one. Consequently, template distributors will need only to distribute a single simple 1-to-1 template.

This also offers simple color conversion to the palette if desired, though it should be made clear that *the ideal way to do conversion is still with either a third-party tool or by hand*.

Config Changes
---
The most significant drawback to implementing this change is the restrictive nature of cross-origin textures in WebGL. Templates _must_ be CORS-enabled lest they fail to be loaded into GL at all. This is resolved by sending all template textures not known to terminate at CORS-enabled sites through a CORS-proxy. In order to facilitate instance owners using different proxies, the following is added to the server config file: 
```
cors {
  proxyBase: "https://cors-anywhere.herokuapp.com"
  proxyParam: ""
}
```
The "proxyBase" option takes a URL which points to a CORS proxy. The default in the reference config is a popular public cors proxy, but instances might choose to use their own server as a proxy if they want.

The "proxyParam" option allows the specification of an HTTP parameter which will be set to the URI of the requested resource. If this option is the empty string - as it is by default - the URI will be joined to the "proxyBase" option with a forward slash before being requested.

Detailed Explaination
---
Functionally, this feature works in two steps:
- Downscaling and conversion
- Stylizing

Each step is accomplished with a separate shader-program. The first one - downscaling - takes the input template as a texture and renders it to an intermediary texture with the size specified by the user-input `tw` (width) parameter. Each pixel of the output is an average of the non-transparent pixels in the associated area in the input. This resulting color is scanned before output and the alpha channel of the intermediary texture is set to a value representing one of:
- The index in the palette the color matches (if any).
- A transparent pixel (no non-transparent pixels were sampled).
- A non-palette color.

The second step - stylizing - converts the intermediary texture into the final template with the active style applied. If the template style is changed, only this step needs to be run to produce a corrected template image. It functions by taking the intermediary texture and a style atlas as input textures. The style atlas needs to contain styles for up-to 255 palette indicies and so is a texture with sides measuring 16 times the size of each palette index style. A simple dotted 1:2 style ('smalldotted' or just 'dotted') is 48 by 48. For producing such a style template, the final output size will be 3 (48 / 16) times the size of the *intermediary* texture.

The palette index is taken from the alpha channel of the intermediary texture and used to address this atlas as follows:
- The first palette entry (index  0) is the top-left atlas entry.
- Each increasing entry is addressed horizontally (1 would be right of 0, 2 right of 1, etc).
- Every 16 entries, the horizontal offset is reset and the vertical offset increases by 1. The puts the sixteenth entry (15) directly below the first (0).
- The special values for a non-palette color (255) and a transparent pixel (254) take up the last and second-last slots respectfully. Since transparent entries are never rendered, putting a style in it's slot has no effect. The non-palette entry falls into the bottom-rightmost slot and can be styled to indicate a wrong color or can be made completely transparent.

Once the appropriate style is selected for drawing a pixel, the color values from the intermediary texture are output using the alpha channel from the style entry. To clarify, this means that **only the alpha channel of style textures is used**.

- [x] Implement template rendering using WebGL
- [x] Implement template downscaling
- [x] Account for pxlsfiddle metadata pixel at top-leftmost pixel of templates
- [x] Implement template palette indexing
- [x] Implement template palette conversion
- [x] Implement template styling
- [x] Allow styling based on palette index
- [x] Allow styling non-palette colors uniquely
- [x] Implement CORS-proxy support
- [x] Implement user front-end for setting template style
- [x] Allow users to bypass WebGL template rendering and use a system similar to the old one
- [x] Store template style settings in localstorage as preferences.


netux (commented):
Don't know much about WebGL, but it worked good so far.

FlyingSixtySix:
Is there anything that can be done to combat this type of scaling artifacting? I'm not seeing it with our current template rendering system; maybe some form of smoothing could be applied? What would be a better way of going about it?

BlankParenthesis:
The template is rendered 1-to-1 by WebGL. This is put into the canvas buffer which is then scaled by the browser. Any differences in scaling are entirely due to how the browser scales `canvas` and `img` elements differently. As far as I'm concerned, they should not be scaled differently so any differences are a browser bug.

BlankParenthesis:
> There might be a property of canvases that controls scaling which is not present on img elements, I'll have a look at that but it might not change anything.
The property I was thinking of is for Context2D which doesn't apply since this uses a WebGL context.

> I'm not seeing it with our current template rendering system
Are you certain of this? You can check by switching the template style between "Use Source Style" and the matching WebGL style. "Use Source Style" uses the old system so you should be able to see a difference if what you say is true.

BlankParenthesis:
Nevermind, should be fixed now.

BlankParenthesis:
Rebased, then merged. Lost a little bit of commit history it seems but the conflicts with #508 should now be resolved.

PR #461, opened 2021-01-08
Set Secure attribute on cookies set over HTTPS

netux:
Undertow defaults cookie's SameSite to "None", but [browsers will soon require websites to also set the Secure attribute when using that SameSite mode](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#samesitenone_requires_secure).

This PR sets Secure and SameSite to "None" over secure connections (requests using HTTPS), and just SameSite to "Lax" on insecure connections, which [is the default on browsers nowadays](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#Lax).


netux:
I was debating whenever to trust the user with what they say is a secure connection (allowing both http and https connections to authenticate fine), or to set the Secure and SameSite attributes depending on some configurable value.

Maybe forcing the more secure alternative by default but making it configurable to trust the user instead is the way to go?

PR #465, opened 2021-01-27
Redact factions in chat and lookups, and usernames in chat purges, during Snip mode

netux:
- Don't send Faction on lookups requested by non-staff users
- Don't send Faction on chat messages during snip mode
- Replace initiator and target usernames with -snip- when sending purges to non-staff users

### DIFF
--- a/src/main/java/space/pxls/App.java
+++ b/src/main/java/space/pxls/App.java
@@ -936,6 +936,10 @@ public class App {
         return virginmap[x + y * width];
     }
 
+    public static boolean getSnipMode() {
+        return getConfig().getBoolean("oauth.snipMode");
+    }
+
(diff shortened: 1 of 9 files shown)


BlankParenthesis:
If this doesn't already cover it, last I checked Discord usernames are sent as -snip- if they are present rather than not being sent.
That might be appropriate to resolve.

netux:
That seems to be on purpose https://github.com/pxlsspace/Pxls/blob/be81dfb5cfafe786f31db0dcd2e3f8f61e109fb3/src/main/java/space/pxls/server/packets/http/Lookup.java#L31
Though honestly, I don't see why that was the case.

I was too slow though, PR already got merged :x.

PR #469, opened 2021-01-30
UI palette tweaks

FlyingSixtySix:
This pull request adds two settings under the **UI Settings** group:
- The ability to toggle the thin scrollbar width on the palette bar
- The ability to override the `white-space` property to allow for stacking

To accomplish the first option, a separate CSS class was made with the following properties:
```css
.thin-scrollbar {
    scrollbar-width: thin;
    scrollbar-color: #aaa transparent;
}
```
The class will be toggled onto the `palette` element.

To accomplish the second option, another CSS class was made with the following property:
```css
.palette-stacking {
    white-space: normal !important;
}
```
When toggled, the palette buttons will overflow onto the next row with no extra limitations for how many rows it can take up. To make the rows look nicer, the top and bottom padding on `#palette` was reduced to `10px` from `15px` and an all-round margin of `5px` was added to the `.palette-color` class.

Demonstration:
ui-palette-tweaks

### DIFF
--- a/resources/public/index.html
+++ b/resources/public/index.html
@@ -318,6 +318,12 @@
                             <span class="label-text">Enable thin scrollbar</span>
                         </label>
                     </div>
+                    <div data-keywords="scrollbar;palette scrolling;palette stack">
+                        <label class="input-group">
+                            <input id="setting-ui-palette-stacking-enable" type="checkbox">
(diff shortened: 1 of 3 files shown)


netux (approved):
lgtm 👍

PR #471, opened 2021-01-31
Wrap WebSocketChannel and optimize sending of some snip mode-aware packets

netux:
This PR adds a new class, `PxlsWebSocketChannel`, which stores an underlying `WebSocketChannel` along with other information that can be linked to that channel, like `User`.

This allows the fix a bug, also done in this PR, where non-authed users wouldn't receive certain chat packets (chat messages and purges).


src/main/java/space/pxls/server/UndertowServer.java
```diff
@@ -222,6 +225,32 @@ public void broadcastToUserPredicate(Object obj, Predicate<User> predicate) {
                 );
     }
 
+    public void broadcastPredicate(Object obj, Predicate<PxlsWebSocketChannel> predicate) {
+        String json = App.getGson().toJson(obj);
+        connections.parallelStream()
```
netux, line 230: I researched what is needed for a parallel stream to work and couldn't find any issues with its usage here.
If something does come up, it can be switched to a regular `.stream()`

src/main/java/space/pxls/server/UndertowServer.java
```diff
@@ -222,6 +225,32 @@ public void broadcastToUserPredicate(Object obj, Predicate<User> predicate) {
...
+        String nonStaffJSON = App.getGson().toJson(nonStaffObj);
+        String staffJSON = App.getGson().toJson(staffObj);
+        broadcastMapped(con -> {
+            boolean sendStaffObject = con.getUser().isPresent() && userCanReceiveStaffBroadcasts.test(con.getUser().get());
+            return sendStaffObject ? staffJSON : nonStaffJSON;
+        });
+    }
+
+    public void broadcastMapped(Function<PxlsWebSocketChannel, String> mapper) {
+        connections.parallelStream()
```
netux, line 245: Ditto about the parallel stream.

src/main/java/space/pxls/server/PxlsWebSocketChannel.java
```diff
@@ -0,0 +1,25 @@
+package space.pxls.server;
+
+import java.util.Optional;
+
+import io.undertow.websockets.core.WebSocketChannel;
+
+import space.pxls.user.User;
+
+public class PxlsWebSocketChannel {
```
netux, line 9: Minor detail, but I was conflicted about the name of this class. Maybe `WebSocketChannelWrapper` or `PxlsWebSocketConnection` are better names for it?

src/main/java/space/pxls/server/PxlsWebSocketChannel.java
```diff
@@ -0,0 +1,25 @@
...
+	private WebSocketChannel underlyingChannel;
```
BlankParenthesis, line 10: The naming of this class seems to indicate that it acts more as a superclass of the WebSocketChannel rather than as a container. There are reasons to favour composition over extension and vice-versa but I figured it was at least worth clarifying that both had been considered.

PR #472, opened 2021-02-06
Shadow-Banned Chat Messages

FlyingSixtySix:
### Database Changes

```sql
ALTER TABLE chat_messages ADD COLUMN shadow_banned BOOL NOT NULL DEFAULT false;
```

### Configuration Changes

```
chat {
    ...
    showShadowBannedMessagesToStaff: true
    ...
}
```

### Goals

This pull request aims to hide new and historical chat messages by shadow-banned users from users without the `chat.history.shadowbanned` and `user.receivestaffbroadcasts` permissions, for history and new messages respectively.  Users with said permissions will see those messages as `.9em`-sized `#F66`-colored text.

### Specifics

- For shadow-banned users, their messages will be processed but not be re-broadcast.  It will be sent back to give the illusion that the message was handled like normal.
- For users without the permissions, no packets will be received regarding new messages from shadow-banned users.
- For users with the permission, messages will have the `data-shadow-banned` element attribute and appear as `.9em`-sized `#F66`-colored text.

### Screenshots

Perspective of a shadow-banned user:

Perspective of a user with the necessary permissions:

### DIFF
--- a/src/main/java/space/pxls/data/Database.java
+++ b/src/main/java/space/pxls/data/Database.java
@@ -1236,7 +1236,7 @@ public class Database {
      * @param includeShadowBanned Whether or not to include shadow-banned messages.
      * @return The retrieved {@link DBChatMessage}s. The length is determined by the {@link ResultSet} size.
      */
-    public List<ChatMessage> getlastXMessagesForSocket(int x, boolean includePurged, boolean includeShadowBanned, boolean ignoreFilter) {
+    public List<ChatMessage> getLastXMessagesForSocket(int x, boolean includePurged, boolean includeShadowBanned, boolean ignoreFilter) {
         DBChatMessage[] fromDB = getLastXMessages(x, includePurged, includeShadowBanned);
         List<ChatMessage> toReturn = new ArrayList<>();
         for (DBChatMessage dbChatMessage : fromDB) {
--- a/src/main/java/space/pxls/data/Database.java
+++ b/src/main/java/space/pxls/data/Database.java
@@ -169,7 +169,8 @@ public class Database {
                     "filtered VARCHAR(2048) NOT NULL DEFAULT ''," +
                     "purged BOOL NOT NULL DEFAULT false," +
                     "purged_by INT," +
-                    "purge_reason TEXT)")
+                    "purge_reason TEXT," +
+                    "shadow_banned BOOL NOT NULL DEFAULT false)")
                     .execute();
             // chat_reports
             handle.createUpdate("CREATE TABLE IF NOT EXISTS chat_reports (" +
@@ -1152,14 +1153,16 @@ public class Database {
      * @param sent The chat message's creation epoch.
      * @param content The chat contents.
      * @param filtered The filtered chat contents.
+     * @param shadowBanned Whether or not the user sending the message is shadow-banned.
      * @return The new chat message's ID.
      */
-    public Integer createChatMessage(int authorID, long sent, String content, String filtered) {
-        return jdbi.withHandle(handle -> handle.createUpdate("INSERT INTO chat_messages (author, sent, content, filtered) VALUES (:author, :sent, :content, :filtered)")
+    public Integer createChatMessage(int authorID, long sent, String content, String filtered, boolean shadowBanned) {
+        return jdbi.withHandle(handle -> handle.createUpdate("INSERT INTO chat_messages (author, sent, content, filtered, shadow_banned) VALUES (:author, :sent, :content, :filtered, :shadow_banned)")
                 .bind("author", authorID)
                 .bind("sent", sent)
                 .bind("content", content)
                 .bind("filtered", filtered)
+                .bind("shadow_banned", shadowBanned)
                 .executeAndReturnGeneratedKeys("id")
                     .mapTo(Integer.TYPE)
                     .first());
@@ -1170,10 +1173,11 @@ public class Database {
      * @param sent The chat message's creation epoch.
      * @param content The chat contents.
      * @param filtered The filtered chat contents.
+     * @param shadowBanned Whether or not the user sending the message is shadow-banned.
      * @return The new chat message's ID.
      */
-    public Integer createChatMessage(User author, long sent, String content, String filtered) {
-        return createChatMessage(author == null ? -1 : author.getId(), sent / 1000L, content, filtered);
+    public Integer createChatMessage(User author, long sent, String content, String filtered, boolean shadowBanned) {
+        return createChatMessage(author == null ? -1 : author.getId(), sent / 1000L, content, filtered, shadowBanned);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


src/main/java/space/pxls/server/PacketHandler.java
```diff
@@ -420,12 +420,21 @@ public void handleChatMessage(WebSocketChannel channel, User user, ClientChatMes
...
+                var staffPacket = barePacket;
+                if (user.isShadowBanned()) {
+                    // To the user, it looks like their message was sent successfully.
+                    server.send(user, userPacket.asShadowBanned());
+                    // To other users, nothing was sent.
+                    userPacket = null;
+                    // To staff, if enabled in the config, they will be the only ones to also get the message.
+                    staffPacket = App.getConfig().getBoolean("chat.showShadowBannedMessagesToStaff") ? staffPacket : null;
+                }
+                server.broadcastSeparateForStaff(userPacket, staffPacket);
```
netux, line 437: Maybe do this is if both `userPacket` and `staffPacket` aren't `null`?

src/main/java/space/pxls/data/Database.java
```diff
@@ -1170,10 +1173,11 @@ public Integer createChatMessage(int authorID, long sent, String content, String
      * @param sent The chat message's creation epoch.
      * @param content The chat contents.
      * @param filtered The filtered chat contents.
+     * @param shadowBanned Whether or not the user sending the message is shadow-banned.
      * @return The new chat message's ID.
      */
-    public Integer createChatMessage(User author, long sent, String content, String filtered) {
-        return createChatMessage(author == null ? -1 : author.getId(), sent / 1000L, content, filtered);
+    public Integer createChatMessage(User author, long sent, String content, String filtered, boolean shadowBanned) {
+        return createChatMessage(author == null ? -1 : author.getId(), sent / 1000L, content, filtered, shadowBanned);
```
netux, line 1180:
```suggestion
        return createChatMessage(author == null ? -1 : author.getId(), sent / 1000L, content, filtered, author == null ? false : author.isShadowBanned());
```
(and removing the `boolean shadowBanned` argument)

src/main/java/space/pxls/server/PacketHandler.java
```diff
@@ -384,7 +384,7 @@ public void handleChatbanState(WebSocketChannel channel, User user, ClientChatba
     }
 
     public void handleChatHistory(WebSocketChannel channel, User user, ClientChatHistory clientChatHistory) {
-        server.send(channel, new ServerChatHistory(App.getDatabase().getlastXMessagesForSocket(100, user.hasPermission("chat.history.purged"), false)));
+        server.send(channel, new ServerChatHistory(App.getDatabase().getlastXMessagesForSocket(100, user.hasPermission("chat.history.purged"), user.hasPermission("chat.history.shadowbanned"), false)));
```
netux, line 387: When the shadow-banned user refreshes, they won't see their own messages.
This shouldn't be much of a problem considering the same happens for pixels on the canvas, but since we are actually storing the messages here its doable.

src/main/java/space/pxls/data/Database.java
```diff
@@ -1229,17 +1234,19 @@ public DBChatMessage getChatMessageByID(int id) {
      * @param x The amount of chat messages to retrieve.
      * @param includePurged Whether or not to include purged messages.
      * @param ignoreFilter Whether or not the chat filter should apply to messages being returned.
+     * @param includeShadowBanned Whether or not to include shadow-banned messages.
      * @return The retrieved {@link DBChatMessage}s. The length is determined by the {@link ResultSet} size.
      */
-    public List<ChatMessage> getlastXMessagesForSocket(int x, boolean includePurged, boolean ignoreFilter) {
-        DBChatMessage[] fromDB = getLastXMessages(x, includePurged);
+    public List<ChatMessage> getlastXMessagesForSocket(int x, boolean includePurged, boolean includeShadowBanned, boolean ignoreFilter) {
```
netux, line 1240: Can we take this as a chance to fix that `l` in "Last" being lowercase? 😖

netux (changes requested):
Looks good. I have some minor comments about it though.

src/main/java/space/pxls/server/PacketHandler.java
FlyingSixtySix: I was initially hoping this would get addressed in another pull request fixing the shadow-banned pixels issue, but given the opportunity I can take care of it here.  What would be the best way to pass in context regarding the requesting user's shadow-ban status?  If I pass in the `user` from `handleChatHistory`, that would knock out both permission check parameters, but looking right above I can see that `ServerChatbanState` takes three parameters, all of them being context from a `User` object.  If replacing the `includePurged` and `includeShadowBanned` parameters with a `User` is fine here, I'll tackle `ServerChatbanState` and whatever else uses the same style in another commit outside this pull request.

netux: 🤔 how about always fetching shadow-banned messages, but filtering them out in `handleChatHistory`?
Otherwise you can add an `int requesterUserID` argument to `getLastXMessageForSocket` for which you always fetch their messages.

netux: If it's too complicated we can work on it in another PR.

netux (approved):

src/main/java/space/pxls/server/PacketHandler.java
FlyingSixtySix: I'll tackle it in `handleChatHistory` - that seems most straight-forward.

src/main/java/space/pxls/server/PacketHandler.java
```diff
@@ -384,7 +385,16 @@ public void handleChatbanState(WebSocketChannel channel, User user, ClientChatba
...
-        server.send(channel, new ServerChatHistory(App.getDatabase().getLastXMessagesForSocket(100, user.hasPermission("chat.history.purged"), user.hasPermission("chat.history.shadowbanned"), false)));
+        var messages = App.getDatabase().getLastXMessagesForSocket(100, user.hasPermission("chat.history.purged"), false);
+        var filtered = messages.stream()
+                .peek(message -> {
+                    // If the user requesting the history is shadow-banned, they should receive all messages, but with
+                    // their own shadow-banned messages NOT marked as shadow-banned. This gives the illusion that they
+                    // sent messages normally.
+                    if (user.isShadowBanned() && message.authorIsShadowBanned) message.authorIsShadowBanned = null;
+                })
+                .collect(Collectors.toList());
```
netux, line 396: Forgot to filter out shadow banned messages depending on the `chat.history.shadowbanned` perm.

netux (changes requested):

src/main/java/space/pxls/server/PacketHandler.java
FlyingSixtySix: I tried thinking of a better way to filter it out based on permission, but c4bfc54 is all I could come up with.  If you have any suggestions on it, please let me know!

FlyingSixtySix:
Superseded by #472.

PR #474, opened 2021-02-09
Use pebble templating for index

BlankParenthesis:
This changes the config slightly: faq and info locations no longer specify if they are a resource or a file and require no such prefixes.

This change was made because using templating for html files is a pre-requisite to automated localization, which seems to be a topic of interest right now.


FlyingSixtySix:
Merging #475 instead.

PR #477, opened 2021-02-14
Change 50x Zoom Limit to Min/Max Inputs

FlyingSixtySix:
### Goals

This pull request aims to replace the "Limit zoom to 50x" option with a couple of number input boxes for custom minimum and maximum scale limits.

### Specifics

- The default minimum and maximum scales are 0.5 and 50 respectively. These are also the placeholder values.
- Users may enter any numerical input, which allows for greater freedom but may lead to bugs. For example, if the user inputs a maximum value less than the minimum value, the canvas coordinates will turn `NaN` and have unexpected behavior.

### Screenshots / Videos

https://user-images.githubusercontent.com/19217244/107891680-7c4d2800-6f20-11eb-9b06-b009c161074a.mp4

### DIFF
--- a/resources/public/index.html
+++ b/resources/public/index.html
@@ -508,11 +508,11 @@
                     <div data-keywords="zooming limit;scrolling limit;mousewheel;zooming minimum;zooming maximum">
                         <label class="input-group">
                             <span class="label-text">Minimum scale:</span>
-                            <input type="number" placeholder="0.5" step="any" min="0" max="50" class="template-coords small-input" id="setting-board-zoom-limit-minimum">
+                            <input type="number" placeholder="0.5" step="any" class="template-coords small-input" id="setting-board-zoom-limit-minimum">
                         </label>
                         <label class="input-group">
                             <span class="label-text">Maximum scale:</span>
-                            <input type="number" placeholder="50" step="any" min="0" max="50" class="template-coords small-input" id="setting-board-zoom-limit-maximum">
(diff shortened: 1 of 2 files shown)


BlankParenthesis:
This looks good at a glance. I do wonder though if it might make more sense to have the inputs represent the exponent of zoom rather than the actual value. This would mean changes to these values manifest as perceptively linear changes rather than logarithmic ones - albeit at the cost of a mismatch between the value and the query parameter.

FlyingSixtySix:
> This looks good at a glance. I do wonder though if it might make more sense to have the inputs represent the exponent of zoom rather than the actual value. This would mean changes to these values manifest as perceptively linear changes rather than logarithmic ones - albeit at the cost of a mismatch between the value and the query parameter.

I'm not sure I see the point of that.

BlankParenthesis:
Thinking more carefully about it, I don't think it my proposition would actually be more intuitive as I thought it would be. The idea is that a user limiting their zoom to a value - suppose 100 - might decide they want to be able to zoom twice as much. To do this, they enter 200 into the input. If they expected that the number of zoom steps from 1 to the maximum would double, they would be very surprised to find it changed only by one or two. If they intended that they would now be able to zoom in to a point where their effective view dimensions were halved, they would find that exactly what they expected happened.

Personally I think the latter expectation is slightly more reasonable, meaning no change is necessary.

PR #500, opened 2021-05-16
Prevent repeated key events from toggling overlays

BlankParenthesis:
This prevents the flashing effect that would previously be triggered by holding down either of the overlay toggle hotkeys. This is done here by only toggling the overlay once per keypress, ignoring repeated events until the key has been unpressed.


FlyingSixtySix:
Hacky. I like it.

PR #511, opened 2021-07-01
Add configurable chat echo via webhooks

BlankParenthesis:
This probably only supports Discord's webhooks. This is a nicer solution than setting up an external bot to connect to the server, since websocket connections require constant maintenance via pings and such which can be unreliable if not handled well.

Information echoed:
- [X] Message content
- [X] Author name (links to profile)
- [X] Timestamp
- [X] Color (probably crashes on special colors now that I think about it…)
- [ ] Badges
- [x] Faction tags
- [x] Embeded links (didn't test that…)
- [ ] Server messages


FlyingSixtySix:
I want to include an emoji property for text-based representations of badges, but that's out of the scope of this PR. Server messages can be added in later.

As for embedded links, I escaped `[` in `relayChatMessageToWebhooks` so trying to make a markdown link won't work.

PR #517, opened 2021-08-15
Minor localization esthetic fix

Seon82:
Switched to non-breaking spaces before interrogation marks in the French localization files, as [recommended by French typographic codes](https://fr.wikipedia.org/wiki/Espace_fine_ins%C3%A9cable). 

Normal spaces would cause the interrogation mark to get isolated on its own line in the FAQ for certain mobile layouts.

### DIFF
--- a/po/Localization_fr.po
+++ b/po/Localization_fr.po
@@ -987,7 +987,7 @@ msgstr "FAQ"
 
 #: resources/public/pebble_templates/faq.html:7:20
 msgid "Why is the cooldown as long as it is?"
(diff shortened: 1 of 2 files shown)


BlankParenthesis:
This looks correct to me. I do wonder though if it might be better to use the HTML `&nbsp;` for clarity.

BlankParenthesis:
Then again, I suppose not all translated strings can be assumed to output in HTML. Leaving it as the character directly or using an escaped character code might be better choices for consistency.

Seon82:
I agree with the second comment, and also feel like getting translators to write HTML probably isn't the right approach.

PR #523, opened 2021-10-17
Add a 'synthwave' theme and donator color

GrayTurtles:
This adds a synthwave/neon-styled theme and a donator gradient color to go with it.
It uses dark background colors and bright blue/pink colors for the text with some glow effects using CSS shadow properties. It also adds colors to the chat role badges.

It's probably easier to show all the changes with screenshots:

_main interface:_

_"Donator Synthwave" color gradient:_

_some settings:_

_chat lookup:_

_panel trigger with a notification + ping:_

_colored role badges:_


GrayTurtles:
I was thinking of making CSS variables for the shadows but there aren't that many shadows that are re-used and it would add more lines to this already way-to-big theme, tho it would make things a little easier to read and configure.

And for the colored badges, I wasn't sure either, another way to make them would be to add a "color" parameter to the role.conf but I wasn't sure how to do that so I went the easy way.

BlankParenthesis:
> add a "color" parameter to the roles.conf

This was my thought too. It's something that can be done later though, so I can see this being accepted first.

On the whole, this looks good to me 👍.

PR #525, opened 2021-11-02
Donator Colour cleanup

Iceydwagon2:
Gray triggering me with that extra line


FlyingSixtySix:
While I don't think this is specific to this PR, something I noticed was that the colors will abruptly change at the end of the gradient. For example:

PR #528, opened 2021-11-12
Use identity server for user authentication

### LINKED ISSUE #482: OAuth Support
Allowing third party pxls tools to "log in with pxls" and gain access to certain data from pxls with permission from the user (yes this would be OAuth for a site that uses OAuth to log in...).
This could allow for verification of account ownership or a few other things like third party programs getting pixel data with the consent of the user it's getting the pixel data of.

BlankParenthesis:
This PR aims to remove user authentication from pxls and offload it to an identity server. This should vastly simplify the role the board server plays in user management.

Identity servers perform the task of user management better than pxls currently does, offering some of the following benefits:
- Better user signup and account management support.
- Better moderation support for managing other user accounts.
- More federation and external provider choices.
- Support for external applications verifying users (using OAuth, closes #482).
- Improved security standards.

Summary of changes
---
*Some changes are improvements, some are neutral, and some a regressions.*
- Users table has dropped field `discord_name` 
- Users table has new field `sub VARCHAR(255) UNIQUE NOT NULL`
- New table `user_links`
- Dropped table `user_logins`
- ~~Dropped~~ **Modified** table `sessions`
- ~~Login sessions no longer survive restart~~
- ~~Usernames can no longer be reset by the pxls server~~
- Usernames **initially** match the id server preferred username
- Discord account is now always shown if connected
- Users can no longer specify Discord username themselves
- Server accepts `Authorization: Bearer …` header for authorization

Required database changes
---
While the server no longer uses the `user_logins` table, dropping it is optional and may be done at leisure.

Altering the `users` table is required, however. The following SQL will do this.
```sql
ALTER TABLE users DROP COLUMN discord_name;
ALTER TABLE users ADD COLUMN sub VARCHAR(255);
-- before running the following, set `sub` for every user.
ALTER TABLE users ALTER COLUMN sub SET UNIQUE NOT NULL;
ALTER TABLE users ADD CONSTRAINT users_sub_key UNIQUE (sub);
```

Required configuration changes
---
The contents of the `oauth` config node have been nearly entirely replaced and the node itself renamed to just `auth`. `useIp` and `snipMode` should function as normal. The remaining values should be configured as follows:
- ~~`callback` should be a url pointing to `/callback` on your site (for example, http://pxls.space/callback)~~
- `issuer` should be a url pointing to the identity server issuer root.
- `client` should be the name of the client configured for use by the pxls server on the identity server.
- `secret` should be empty if the client is a public one or the secret for that client otherwise.

Additionally, the root level `host` value should now be a url pointing the client, scheme and port included. For example, Pxls would use `"https://pxls.space/"`. A local development environment might use `"http://localhost:3000"`.

There is also a new `auth.devmode` node which will disable ssl verification for auth requests, allowing the use of a locally hosted self-signed identity server.

Required other changes
---
Since the `user_logins` table is dropped, it will need to be migrated beforehand. This is identity-server dependent.

Identity server configuration
---
The identity server itself requires some specific configuration.

 In order for pxls to see a user's connected accounts, the ID server should offer a `connected_accounts` scope. This claim should add an `accounts` claim to the user's generated tokens which contains a list of objects with the following properties:
- `userid`: the user's service ID
- `userName`: the user's service username
- `identityProvider`: the name of the service

Pxls also requires access to the `profile` scope to access the `preferred_username` claim.

The ID server should allow a redirect_uri of `{site}/auth?client_name=OidcClient`.


BlankParenthesis:
Finally taking another look at this, I've resolved the merge conflicts for now and updated things to ensure they function. This will also need a separate PR for the client as the changes go hand-in-hand.

While looking it over, here are my observations of what this requires to be "ready":
- [x] Using a more permanent token store rather than the in memory one.
- [X] Deleting the now unused auth providers.
- [ ] ~~Optimizing account link syncing (and more caching of resources in general).~~ **Requires revisiting UserManager and its cache, which is far out of scope for this PR.**
- [x] Actually reload the configuration on reload if possible.
- [x] Maybe reverting the changes about username updates: from experience with the new rust server (which does use openid already) allowing the username to diverge from `preferred_username` is actually a useful and sensible behavior to have.

For further work, actually resolving #482 will probably require adding some new grants to control what endpoints can be accessed by a third-party application, since ubiquitous user-level access is a security risk.

BlankParenthesis:
This seems almost good to go, the client changes in https://github.com/pxlsspace/pxls-web/pull/65 will need to be merged at the same time or things will break.

PR #535, opened 2021-12-01
Add three name gradients for Donators: Mines, Eggplant, and Banana

Avery-Dunktastic:
It's pretty much just that. The next time I get the opportunity to personally test new gradients (Globe, Copper, Trees), I will update this.


Avery-Dunktastic:
Again, I'm going to see if I can add three more gradients, all of which will be general themes I don't have any real bias towards.
If you'd rather I create the other three gradients before submitting this, do let me know.

PR #544, opened 2021-12-06
Fix the canvas getting locked when pressing 'R' without template on a new tab

GrayTurtles:
This was due to the template always being displayed on init so if there was not actually a template, the coords were undefined which locks the canvas.


BlankParenthesis:
Looks good to me

PR #566, opened 2022-03-19
Add ability to use local files for templates

FlyingSixtySix:
This adds two new functionalities: dragging and dropping images from the host PC onto Pxls, and browsing to files from the template settings. Both will convert the supplied image to base64 and use it as the template URL.

Demonstration video:

https://user-images.githubusercontent.com/19217244/159108399-4e71677c-d272-4ca5-96f9-e3cd4004b595.mp4

### DIFF
--- a/po/Localization.po
+++ b/po/Localization.po
@@ -76,6 +76,10 @@ msgid "Username:"
 msgstr "Username:"
 
 #: resources/public/pebble_templates/index.html
+msgid "Drag and drop template image"
+msgstr "Drag and drop template image"
+
+#: resources/public/pebble_templates/index.html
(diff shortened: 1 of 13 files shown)


BlankParenthesis:
Looks great! Just a couple of issues I spotted:

If you drag a file onto the page then drop it outside of the page into another program, the drop notification remains until you actually drop a file onto the page.

The other issue is that of file type filtering — just checking that the mimetype decodes to `image/*` is insufficient. Several filetypes are recognised as images but not necessarily supported by the browser. For example, `image/x-xcf` will accept the drop and then fail once the page actually tries to parse the contents in the template section itself.

PR #587, opened 2022-04-12
Reply feature

int3nse-git:
### Adds the ability to reply to messages in the chat! 

* Client sides finds message by ID, so no risk of loading old or purged messages
* Acts as a mention to the user being replied to
* Ability to click on message previews to jump to them
* Message previews not in client's message history listed as "Message couldn't be found"
* Message previews that refer to messages that have been purged will be replaced by "Message was purged"
* Users that have permission to see purged messages will instead see the preview styled the same as purged messages would be
* ^and the same thing for shadow banned messages, in the weird case that one is replied to (ex: by the shadow banned user)

# Required database changes
* replying_to_id - default -1 to avoid problems with a 0 indexed message, stores the id of the message that is being replied to
* reply_should_mention - default doesn't really matter, stores if the author of the message being replied to should be mentioned
```sql
ALTER TABLE chat_messages ADD COLUMN replying_to_id BIGINT NOT NULL DEFAULT -1;
ALTER TABLE chat_messages ADD COLUMN reply_should_mention BOOL NOT NULL DEFAULT true;
```

### DIFF
--- a/po/Localization.po
+++ b/po/Localization.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Pxls\n"
-"POT-Creation-Date: 2022-03-24T16:29:23.229Z\n"
+"POT-Creation-Date: 2022-04-09T15:04:32.029Z\n"
 "Language: en\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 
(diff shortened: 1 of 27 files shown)


int3nse-git:
Oh yeah, and I did a quick look over to make sure that it looks OK on every theme, but of course it's only a quick edit if anything needs improvement

int3nse-git:
Edited "Required database changes" in original pr message

int3nse-git:
Now this should *actually* be ready to merge.

int3nse-git:
Now this should *actually* be ready to merge.

int3nse-git:
> "Now this should _actually_ be ready to merge."

Once again wrong, more changes incoming

PR #606, opened 2022-05-14
Switch color remapping to delta EITP

Seon82:
The [delta_eitp](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2124-0-201901-I!!PDF-E.pdf) method for color remapping should provide better results than the current color distance. Below is a comparison of different color distance functions for a more visual argument, the method currently used being luma;

image

NB: I also have an implementation of [redmean](https://en.wikipedia.org/wiki/Color_difference#sRGB) I can add to the PR if you'd rather use that one instead.

### DIFF
--- a/resources/public/include/template.js
+++ b/resources/public/include/template.js
@@ -615,7 +615,7 @@ module.exports.template = (function() {
           }
 
         float diffEITP(vec3 col1, vec3 col2) {
-          return distance(0.5*rgb2ictcp(col1) - 0.5*rgb2ictcp(col2));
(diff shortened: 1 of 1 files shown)


BlankParenthesis:
It might be best to add this as a new conversion type rather than replacing the existing one for compatibility reasons. I'm not against eventually removing the current one, but it should probably stay for now and new conversions should also get appropriate names.

Doxylamin:
PR seems stale and hasn't been updated after being requested to review.

PR #610, opened 2022-07-16
Adding Swedish translations to the site

Temriel:
Should be all of them except those not featured in the document

### DIFF
--- a/po/Localization_sv.po
+++ b/po/Localization_sv.po
@@ -1,0 +1,2576 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: Pxls\n"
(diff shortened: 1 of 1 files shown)


BlankParenthesis:
Look pretty good to me, there are just a couple of issues that should probably be fixed before a merge:
- There are some minor formatting differences when compared to the other po files. This is non-critical since the tools should handle it fine but it's worth pointing out.
- The language code needs to be added on the server and in the client override settings.
- The generation script for converting po into a properties file should be run and that properties file committed.

If you're not comfortable with doing that yourself, @FlyingSixtySix should be able to handle the required changes.

Temriel:
Yeah I don't really know how to do that, so if @FlyingSixtySix could do it, I would be glad!

PR #642, opened 2023-02-23
Bleeped out the n-word

zulujive:
I think this is probably best for this to be enabled by default, there's no reason why the n-word should be spelled out fully like that.


zulujive:
Take that back, this code is flawed, DON'T USE IT

PR #650, opened 2023-04-27
Add setting to disable undo

Nizrabed:
Localizations for this setting still need to be configured.

### DIFF
--- a/resources/public/include/place.js
+++ b/resources/public/include/place.js
@@ -296,16 +296,18 @@ module.exports.place = (function() {
         uiHelper.toggleCaptchaLoading(false);
       });
       socket.on('can_undo', function(data) {
-        document.body.classList.add('undo-visible');
-        self.elements.undo.addClass('open');
-        self.can_undo = true;
-        if (self.undoTimeout !== false) clearTimeout(self.undoTimeout);
-        self.undoTimeout = setTimeout(function() {
-          document.body.classList.remove('undo-visible');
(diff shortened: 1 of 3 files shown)


BlankParenthesis:
To my understanding, this stems mostly from a desire to disable the display of the UI button. While this does that, it also looks like it does exactly as the title claims: **disables undo entirely**. This is not the same thing, as undo could still be used with the hotkey even if the button were hidden. I think most people who want this change would want the undo button to still work.

That said, my personal opinion is against merging this: I worry it sets the example that any piece of the UI should be optional and will end up having a setting. Pxls already has perhaps too many configuration options as-is, making the more useful ones harder to access.

I think if we want to go down the route of user UI  tuning, a more advanced theme engine which allows granular changes to be enabled and disabled is a better choice.

FlyingSixtySix:
I agree with @BlankParenthesis; it'd be best to implement a better theming engine to allow for more granular adjustments. I'll leave this open for the time being for reference.

FurryMasterGithub:
Devs, you can add this small function before whole pxls will be recoded and updated (never maybe), accept this very simple function and you can very simple convert it in future to new engine
Accept this right now better than waiting slow big update

BlankParenthesis:
For those that want to hide the UI button, the preferred way for now is with some custom user CSS. Something as simple as this should do the trick:

```css
#undo {
    display: none;
}
```

For those that want to disable undo on the server side: you can set the undo window in the config to 0

PR #659, opened 2023-07-10
Twitch Authentication

LordKyubae:
Added TwichAuthService

### DIFF
--- a/src/main/java/space/pxls/auth/TwitchAuthService.java
+++ b/src/main/java/space/pxls/auth/TwitchAuthService.java
@@ -49,7 +49,7 @@ public class TwitchAuthService extends AuthService {
             return null;
         }
 
-        return httpResponse.getBody().getObject().getJSONArray("data").getJSONObject(0).getString("login");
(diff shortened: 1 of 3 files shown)


BlankParenthesis:
This looks good at a glance. The only issue I see immediately is that the comment in the config file is wrong.

src/main/java/space/pxls/auth/TwitchAuthService.java
```diff
@@ -0,0 +1,66 @@
...
+                .header("User-Agent", "pxls.space")
+                .header("Authorization", "Bearer " + token)
+                .header("Client-Id", App.getConfig().getString("oauth.twitch.key"))
+                .asJson();
+
+        if (!httpResponse.isSuccess()) {
+            return null;
+        }
+
+        return httpResponse.getBody().getObject().getJSONArray("data").getJSONObject(0).getString("login");
```
Lordshinjo, line 52: `login` is not a stable identifier since it can be changed every 60 days (and can even be reused across users), I believe `id` should be used instead.

PR #663, opened 2023-07-15
Update Localization_sv.po

Temriel:
Added some missing chat stuff and replaced all instances of "moderatörer" with "moderatorer". 
Finally changed all instances of "mall" with "template".

// Should note that all rules are still yet to be translated due to them not being available in the .pot file. They do exist in the #translations thread in the pxls.space discord server however, where they are fully translated and ready to be used. I just don't know how to do that personally.

### DIFF
--- a/po/Localization_sv.po
+++ b/po/Localization_sv.po
@@ -642,15 +642,15 @@ msgstr "Interfas"
 
 #: resources/public/pebble_templates/index.html
 msgid "enable;disable"
-msgstr ""
+msgstr "aktivera;avaktivera"
 
 #: resources/public/pebble_templates/index.html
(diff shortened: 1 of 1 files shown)


BlankParenthesis:
> …all rules are still yet to be translated due to them not being available in the .pot file

Assuming you are talking about the pxls.space specific ones, they appear only on site.

> I just don't know how to do that personally.

The changes can be made to the running instance of the site, which is a task I'll have to do manually. This is obviously not an ideal situation and it's something that all instances have to deal with. Perhaps someday there will be a more ideal solution, but the design of such a system seems nontrivial and eludes me in the moment.

PR #672, opened 2023-07-26
Fix regex for discord name update

Nizrabed:
Discord's new usernames cannot include uppercase letters.


BlankParenthesis:
Since this seems to attempt to fix all remaining locations, it seems to be missing the placeholder of the signup field (`Username#1234`).

PR #682, opened 2023-09-01
Improve the info bubble

Nizrabed:
Add some margin to pixel count and username.
Make the keys of pixel counts bold.
Hide the placeable count 
while the cooldown is visible, and make the placeable count visible again otherwise.

Current:
image

Improved:
image

### DIFF
--- a/resources/public/include/timer.js
+++ b/resources/public/include/timer.js
@@ -8,6 +8,7 @@ module.exports.timer = (function() {
   const self = {
     elements: {
       palette: $('#palette'),
+      placement_info: $('#placement-info'),
       timer_container: $('#cooldown'),
       timer_countdown: $('#cooldown-timer'),
       timer_chat: $('#txtMobileChatCooldown')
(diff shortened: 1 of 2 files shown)


BlankParenthesis:
Two small issues with this: 

Firstly, the icons don't appear to align nicely for me — they appear slightly too high:

icon alignment current

They can be made to display more inline with the text by setting `#bubble-grid` to have `align-items: center;`:

icon alignment proposed

An issue with this is potentially the stats icon ending up between then two separate statistics shown, but this could also work to the advantage of making it more obvious the icon relates to both.

Secondly, the coordinates changing can induce size changes in the bubble width which can be rather distracting. This is pretty minor, so it's not a requirement for getting this merged, but perhaps it would be worth considering fixing the minimum width of this element in some way to avoid this. The timer doesn't seem to have this same problem, but I foresee a similar issue occurring there so I wouldn't be opposed to something similar for that.

Nizrabed:
I'll try to fix those problems before canvas reset
 Thanks for letting me know 👍

PR #687, opened 2023-10-14
Check if we can access the template before using the CORS proxy

april83c:
When loading a template, the current behavior is to check if the template URL's domain is on the safe hosts list (a list of websites that serve CORS headers that allow Pxls to load that template), and if it's not on the safe hosts list, use the CORS proxy.

This means that if a template that *isn't* on the safe hosts list serves CORS headers that allow Pxls to load it, it will *still* get proxied through the CORS proxy, even though we could load it directly.

This pull request adds a check where it sends a HEAD request to the template URL to see if it's allowed to load it, and if it is, then it doesn't use the CORS proxy.

### DIFF
--- a/resources/public/include/template.js
+++ b/resources/public/include/template.js
@@ -275,21 +275,23 @@ module.exports.template = (function() {
       }
 
       if (self.elements.styleImage.prop('src') !== self.options.style) {
-        fetch(self.cors(self.options.style), {
-          method: 'GET',
-          credentials: 'omit'
-        }).then(response => {
-          if (response.ok) {
-            return response.blob();
-          } else {
-            throw new Error(`HTTP ${response.status} ${response.statusText}`);
-          }
-        }).then(blob => {
-          const reader = new FileReader();
-          reader.onload = () => {
-            self.elements.styleImage.attr('src', reader.result);
-          };
-          reader.readAsDataURL(blob);
+        self.cors(self.options.style).then(url => {
+          fetch(url, {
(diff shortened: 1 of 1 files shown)


april83c:
For testing purposes, I set up some test templates:
- https://aprl.moe/test-template.png has no CORS headers (should fail and use CORS proxy)
- https://aprl.moe/test-template-cors.png has `Access-Control-Allow-Origin` set to `*` (should succeed and bypass the CORS proxy)

BlankParenthesis:
I have tested this and it appears to work well despite one minor thing: the `cors` function is used in more than the one place it has been edited to be `await`ed. This means that templates fail to load when using webgl template render.

Changing the request on line 278 of `template.js` to something like `self.cors(self.options.style).then((url) => fetch(url, {` should resolve this without needing to make the whole `_update` function also async (and propagating that).

april83c:
> I have tested this and it appears to work well despite one minor thing: the `cors` function is used in more than the one place it has been edited to be `await`ed. This means that templates fail to load when using webgl template render.
> 
> Changing the request on line 278 of `template.js` to something like `self.cors(self.options.style).then((url) => fetch(url, {` should resolve this without needing to make the whole `_update` function also async (and propagating that).

Ah, sorry, didn't catch that. I'll fix it in a bit.

april83c:
Works fine after this latest commit afaict

BlankParenthesis:
Seems to be good to me :+1:

PR #702, opened 2024-06-18
7TV Integration + Move chat history to GET /chat/history

VanillaSixtySix:
This PR adds the server-side component to 7TV emote integration. `chat.emoteSet7TV` must be set to an emote set ID.


Mubelotix:
You forgot to add config values in the reference

PR #709, opened 2024-09-20
Update and add python scripts

Mubelotix:
I recently hosted an event using your codebase (thanks again!). I have found it to be useful to be able to filter the board image by only including specific users. Here is a script that does that from pixels.log


Mubelotix:
Superseded by #714

PR #716, opened 2025-10-19
Streamline chat gradient colors

peyrov:
Makes chat gradient colors configurable from the Pxls configuration file.


resources/roles-reference.conf
```diff
@@ -85,7 +102,7 @@ staff {
     chat.lookup
     chat.history.purged
     chat.history.shadowbanned
-    chat.usercolor.rainbow
+    chat.usercolor.gradient.*
```
BlankParenthesis: This value should either be quoted or changed to just `chat.usercolor.gradient` as the config parser does not like unquoted asterisk characters. It may also be worth considering `chat.usercolor.gradient.rainbow` instead for consistency.

resources/roles-reference.conf
```diff
@@ -46,7 +46,24 @@ donator {
     }
   ]
   permissions: [
-    "chat.usercolor.donator.*"
```
BlankParenthesis, line 49: Simply changing this to `"chat.usercolor.gradient.*"` seems like a more sensible choice than manually listing every gradient, given the wildcard is a feature and already used for the staff role.

docs/roles.md
```diff
@@ -69,10 +69,9 @@ roleID {
 | `chat.purge` | `/admin/chatPurge` | Purge (multiple) chat messages | staff |
 | `chat.report` | `/reportChat` | Report chat messages | user |
 | `chat.send` | | Send chat messages | user |
-| `chat.usercolor.rainbow` | | Ability to use rainbow user color | staff |
-| `chat.usercolor.donator`, `chat.usercolor.donator.*` | | Ability to use donator user colors | donator |
-| `chat.usercolor.donator.green` | | Ability to use green donator user color | donator |
-| `chat.usercolor.donator.gray` | | Ability to use green donator user color | donator |
+| `chat.usercolor.gradient`, `chat.usercolor.gradient.*` | | Ability to use all gradient user colors | staff |
+| `chat.usercolor.gradient.green` | | Ability to use green gradient user color | donator |
```
BlankParenthesis: Both this and the entry below seem a bit lacking now, given the other gradients are missing.

Perhaps simply including a generic note would be better:
```| `chat.usercolor.gradient.<name>` | | Ability to use the gradient specified by <name> | donator |```

BlankParenthesis (changes requested):
Just a few things to fixup on this Other than that, this is a great addition :+1:

resources/roles-reference.conf
peyrov: The problem with using the wildcard is that it also grants the permission for the "rainbow" gradient. Previously that gradient was hardcoded under a different permission node so the wildcard would have excluded it, because the gradient is supposed to be staff exclusive.

BlankParenthesis: Ah, yes I had missed that. In that case I suppose this is the best alternative.

PR #718, opened 2025-11-17
Fix race condition on concurrent board access

shuni64:
Pxls sometimes returns corrupted board data when multiple clients are fetching `/boarddata` or one of the other memory-mapped buffers. [This script](https://gist.github.com/shuni64/80b244d62c94275d29e8f3fcf7dc4a5d) very reliably reproduces the issue on my machine, but due to the nature of race conditions some parameters may have to be tweaked if it doesn't work (namely, increase the number of workers and the number of requests).

I believe the cause of this to be that ByteBuffers have an internal cursor/position that gets reset when a new request starts reading the buffer. Since this value is shared by all readers, existing readers start from the beginning even though they're already halfway through the buffer and have sent that data to the client. In theory this may cause all kinds of weird issues, but this is the behavior I've seen manifest on pxls.space.
There is a likely related issue that, after running the above script, causes the returned board data to be corrupted until Pxls is relaunched, but I haven't been able to determine the root cause of that.
Additionally, [ByteBuffers are not thread-safe](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/nio/Buffer.html), meaning that concurrent access must generally be synchronized.

This PR solves the problem by:
- creating new read-only ByteBuffers with a shared underlying buffer but separate cursors
- synchronizing most other board data accesses

There is still a theoretical issue that could cause slight corruption of reads while a pixel is being set because access to the buffer itself is not synchronized, but I think that, in practice, this does not occur at a high enough rate to warrant locking the board during reads.

### DIFF
--- a/src/main/java/space/pxls/App.java
+++ b/src/main/java/space/pxls/App.java
@@ -979,28 +979,33 @@ public class App {
     }
 
     public static ByteBuffer getHeatmapData() {
-        heatmap.rewind();
-        return heatmap;
+        ByteBuffer copy = heatmap.asReadOnlyBuffer();
+        copy.rewind();
+        return copy;
     }
 
     public static ByteBuffer getVirginmapData() {
-        virginmap.rewind();
-        return virginmap;
+        ByteBuffer copy = virginmap.asReadOnlyBuffer();
+        copy.rewind();
+        return copy;
     }
 
     public static ByteBuffer getPlacemapData() {
-        placemap.rewind();
-        return placemap;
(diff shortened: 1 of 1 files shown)


BlankParenthesis:
Indeed, while developing [pxls-rs](https://github.com/BlankParenthesis/pxls-rs) I got the growing impression that Pxls probably has a lot of questionable race conditions. While I don't doubt that there are more, thanks for finding and addressing this issue at least :+1:.

I can imagine this will impact performance a little, but hopefully it shouldn't be too bad. If it ends up being a problem, it can be revisited.

